PR #8, opened 2023-12-28
feat: add shared types package

connorlindsey:
Adds a `@knocklabs/types` package with a few shared types. Open to feedback on what other types we want to include there. Looking through the other packages, most of the types are pretty specific to each package, so it depends on where we want to draw the line of what gets put in types and what stays local to each package.


kylemcd (approved):
Nice! Definitely would be cool to find a way to pull in some of the types from our BE automatically here.

PR #15, opened 2024-01-17
feat(kno-5010): Add SlackKit and slack example apps

meryldakin:
This PR has been a long running branch that adds the entire SlackKit plus 2 example apps.


examples/slack-connect-example/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
```
connorlindsey, line 1: Nit: we should use the shared eslint config here like the existing nextjs-example

connorlindsey, line 1: Nit: we should use the shared eslint config here like the existing nextjs-example

examples/slack-connect-example/.gitignore
```diff
@@ -0,0 +1,36 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
connorlindsey, line 1: Let's double check, but I don't think this is necessary because there's a gitignore in the root. If we need to, we can extend that one

examples/slack-connect-example/public/favicon.ico
connorlindsey, line 1: Small detail, but I think we should update the examples to use Knock favicons/logos. Could put the assets in a shared folder. Not urgent though and could be something we do in a follow up issue/pr

examples/slack-connect-example/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
```
connorlindsey, line 1: Same comment for the eslint config, we can use the shared tsconfig

packages/react/.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 20.0.0
```
connorlindsey, line 1: Can probably remove this now or update the root `.tool-versions` if needed

packages/react/src/modules/core/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,36 @@
...
+  onColor?: string;
+  onChange?: () => void;
+}
+
+export const Switch: React.FC<SwitchProps> = ({
+  isChecked = false,
+  onColor,
+  onChange,
+}) => {
+  console.log("is Checked", isChecked)
```
connorlindsey, line 15: Nit: remove console log

packages/react/src/modules/core/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const [on, setOn] = useState(isChecked || false);
+
+  const toggleSwitch = () => {
+    if (onChange) {
+      onChange();
+      setOn(!on);
```
connorlindsey, line 21: Nit: could simplify logic to only call `setOn` once (outside of the if statement)

packages/react/src/modules/core/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,36 @@
...
+    } else {
+      setOn(!on);
+    }
+  };
+
+  return (
+    <label className="switch">
```
connorlindsey, line 28: Wdyt about allowing customizing the label/input like by adding `name` or `id` for better accessibility? 

Alternatively, we could reach for something like Radix or Headless UI for high quality primitives out of the box.

packages/react/src/modules/core/components/Switch/Switch.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <input type="checkbox" checked={on} onChange={toggleSwitch} />
+      <span
+        className="slider round"
```
connorlindsey, line 31: I'm not sure the correct answer here, but I think we need some standards around class naming to avoid colliding with end user styles. Maybe prefixing everything with `rnf-` as a lot of components already do.

packages/react/src/modules/slack/components/ConnectToSlackToggle/ConnectToSlackToggle.tsx
```diff
@@ -0,0 +1,70 @@
...
+      user_token: userToken,
+    }),
+    client_id: slackClientId,
+    scope: "chat:write,chat:write.public,channels:read,groups:read",
+  } as Record<string, string>;
+
+  const params = new URLSearchParams(rawParams);
+
+  const handleToggle = () => {
+    if (isChecked) {
```
connorlindsey, line 42: Nit: can be simplified to `setIsChecked(!isChecked)`

packages/react/src/modules/slack/components/ConnectedSlackChannelPicker/ConnectedSlackChannelPicker.tsx
```diff
@@ -0,0 +1,216 @@
...
+export const ConnectedSlackChannelPicker: React.FC<Props> = ({
+  user,
+  tenant,
+  connectionsObject,
+  knockClientId,
+  host = "api.knock.app",
+  userToken,
+  knockSlackChannelId,
+}) => {
+  const knockClient = useMemo(() => {
```
connorlindsey, line 26: Could we use the knock client from `useKnock` (from the `<KnockProvider />`). I think it's safe to assume that provider will be wrapping any usage of the slack components and could simplify these components

packages/react/src/modules/slack/components/ConnectedSlackChannelPicker/ConnectedSlackChannelPicker.tsx
```diff
@@ -0,0 +1,216 @@
...
+
+  const { channels } = data;
+
+  const handleOptionClick = async (channelId: string) => {
+    setIsSettingChannels(true);
+    const channelToUpdate = channels.find(
+      (channel: Channel) => channel.id === channelId,
+    );
+    if (channelToUpdate) {
+      const updatedChannels = channels.map((channel: Channel) =>
```
connorlindsey, line 63: Minor comment: I think we can remove `updatedChannels` and combine the two maps (or chain the filter and 2 maps)

packages/react/src/modules/slack/components/ConnectedSlackChannelPicker/ConnectedSlackChannelPicker.tsx
```diff
@@ -0,0 +1,216 @@
...
+              if (!hasError) {
+                setShowChannelPicker(!showChannelPicker);
+              }
+            }}
+          >
+            {channelsConnectedMessage()}
+            {!hasError && (
+              <span className="icon">
+                {showChannelPicker ? (
+                  <svg
```
connorlindsey, line 142: Nit but it'd be nice to break out these icons into separate components (could have an icons folder) as it'd make this component a bit more concise.

packages/react/src/modules/slack/hooks/useSlackChannels.ts
```diff
@@ -0,0 +1,80 @@
...
+  data: { channels: Channel[] };
+  isLoading: boolean;
+  error: boolean;
+} {
+  const [isLoading, setIsLoading] = useState(true);
+  const [data, setData] = useState<{ channels: Channel[] }>({
+    channels: [],
+  });
+
+  const knockClient = useMemo(() => {
```
connorlindsey, line 36: Same as the picker component, I think we might be able to rely on `useKnock` to get the client.

connorlindsey (commented):
This is looking really good! I think there are some optimizations we can make being in the monorepo and with the new `KnockProvider`, as well as some points we can discuss as a team like handling class names.

packages/react/src/modules/core/components/Switch/Switch.tsx
meryldakin: After talking with Krisna I'm removing the toggle entirely for now, I don't think it's the right time to use it

packages/react/src/modules/slack/components/ConnectedSlackChannelPicker/ConnectedSlackChannelPicker.tsx
meryldakin: okay i think i'll tackle this in another PR as we're probably going to change a lot of the look of this, then will introduce that

package-lock.json
connorlindsey, line 1: We're using yarn in this repo so we should delete this

examples/slack-connect-example/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "jsonwebtoken": "^9.0.0",
```
connorlindsey, line 12: This should define `"@knocklabs/react": "*"` as a dependency and import components there using an absolute path instead of the relative path. With the relative path, this example would break if moved out of the monorepo.

examples/slack-connect-example/pages/index.tsx
```diff
@@ -0,0 +1,163 @@
+import useSetToken from "../hooks/useSetToken";
+import {
+  ConnectToSlackButton,
+  ConnectToSlackContainer,
+  ConnectedSlackChannelPicker,
+  KnockProvider,
+} from "../../../packages/react";
```
connorlindsey, line 7: Should import from `@knocklabs/react` instead of the relative path

connorlindsey (approved):
🥳

PR #21, opened 2024-01-23
feat: update React Native SDK to use new providers and add React Native example

connorlindsey:
- Updates React Native build set up (vite.config)
- Removes NX specific config
- Exports hooks/providers from react-core
- Adds React Native example app and example usage in the package readme


packages/react-native/README.md
```diff
@@ -67,7 +83,7 @@ const YourAppLayout = () => {
     notificationFeed.fetch();
   }, [notificationFeed]);
 
-  return <span>Total unread: {metadata.unread_count}</span>;
+  return <View>Total unread: {metadata.unread_count}</View>;
```
kylemcd, line 86: Text will need to be wrapped in a `react-native` `<Text/>` component, just putting it in a view will cause it to error.

packages/react-native/src/modules/feed/components/NotificationFeedContainer/NotificationFeedContainer.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import { View } from "react-native";
 
-export const KnockFeedContainer: React.FC<React.PropsWithChildren> = ({
+export const NotificationFeedContainer: React.FC<React.PropsWithChildren> = ({
```
kylemcd, line 4: Not apart of the changed code but curious why `React.FC<React.PropsWithChildren>` is here vs defining the prop types directly to the props?

connorlindsey: Do you mean why use `React.PropsWithChildren` instead of creating a props type with a child property? There's not a huge difference but I think using `PropsWithChildren` is more consistent (e.g. if React changes the type children should be).

packages/react-native/README.md
connorlindsey: Ahh nice catch, missed that in the readme

packages/react-native/src/modules/feed/components/NotificationFeedContainer/NotificationFeedContainer.tsx
connorlindsey: Good idea! Yeah, at this point this component is mainly to test that the build is working as expected. In the web react version of this package, the `NotificationFeedContainer` components imports some base styles which we don't have yet. I think for now having it as a plain wrapper is okay.

kylemcd (approved):

cjbell (approved):

PR #28, opened 2024-01-25
chore: run tests in ci

connorlindsey:
Adds basic ci github action to run tests. Note, Turborepo already detects which packages have been changed and will only run tests in updated packages, so no need for `paths-filter` or similar checks.

Screenshot 2024-01-25 at 9 33 56 AM


packages/client/package.json
```diff
@@ -32,7 +32,8 @@
     "lint": "eslint \"src/**/*.+(ts|js|tsx)\"",
     "format": "prettier \"src/**/*.{js,ts,tsx}\" --write",
     "format:check": "prettier \"src/**/*.{js,ts,tsx}\" --check",
-    "test": "vitest",
+    "test": "vitest run",
```
connorlindsey, line 35: Vitest run in watch mode by default locally which conflicts with Turborepo's caching, so we need to specify separate commands.

.github/workflows/cicd.yml
```diff
@@ -0,0 +1,25 @@
+name: CI/CD
+
+on: [push]
```
kylemcd, line 3: Do we want to run this on every "push"? i.e. this would run if you push on a branch that isn't in a PR, resulting in unnecessary runs.

Running it on `pull_request` I think would make more sense, as it would run when opening the PR and any time you push new code to the pr (`synchronize` trigger), here's some docs: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

kylemcd (approved):

PR #29, opened 2024-01-25
feat(kno-5046): Add disconnect and error states

meryldakin:
Add disconnect and error states.

Product demo + code review: https://www.loom.com/share/d826accc9cda412c89d8f882d2ced267

### DIFF
--- a/packages/client/src/clients/slack/index.ts
+++ b/packages/client/src/clients/slack/index.ts
@@ -24,7 +24,17 @@ type GetSlackChannelsInput = {
   knockChannelId: string;
 };
 
-const TENANT_COLLECTION = "$tenants"
+type AuthCheckInput = {
+  tenant: string;
+  knockChannelId: string;
+};
+
+type RevokeAccessTokenInput = {
+  tenant: string;
+  knockChannelId: string;
+};
+
+const TENANT_COLLECTION = "$tenants";
 
 class SlackClient {
   private instance: Knock;
@@ -33,20 +43,40 @@ class SlackClient {
     this.instance = instance;
   }
 
-  async getChannels(params: GetSlackChannelsInput) {
+  async authCheck({ tenant, knockChannelId }: AuthCheckInput) {
     const result = await this.instance.client().makeRequest({
       method: "GET",
-      url: `/v1/providers/slack/channels`,
+      url: `/v1/providers/slack/${knockChannelId}/auth_check`,
       params: {
         access_token_object: {
-          object_id: params.tenant,
+          object_id: tenant,
+          collection: TENANT_COLLECTION,
+        },
+        channel_id: knockChannelId,
+      },
+    });
+
+    return this.handleResponse(result);
+  }
+
+  async getChannels({
+    knockChannelId,
+    tenant,
+    connectionsObject,
+  }: GetSlackChannelsInput) {
+    const result = await this.instance.client().makeRequest({
+      method: "GET",
+      url: `/v1/providers/slack/${knockChannelId}/channels`,
+      params: {
+        access_token_object: {
+          object_id: tenant,
           collection: TENANT_COLLECTION,
         },
         connections_object: {
-          object_id: params.connectionsObject.objectId,
-          collection: params.connectionsObject.collection,
+          object_id: connectionsObject.objectId,
+          collection: connectionsObject.collection,
         },
-        channel_id: params.knockChannelId,
+        channel_id: knockChannelId,
       },
     });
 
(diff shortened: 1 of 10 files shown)


packages/react/src/modules/slack/hooks/useSlackChannels.ts
```diff
@@ -16,13 +15,10 @@ export type ContainerObject = {
...
   tenant: string,
   connectionsObject: ContainerObject,
-
   knockSlackChannelId: string,
-
-  user: User,
   shouldRefetch: boolean,
-  setHasError: (hasError: boolean) => void,
-  hasError: boolean,
+  setErrorMessage: (errorMessage: string | null) => void,
```
connorlindsey, line 20: Nice refactor from the boolean to a message

packages/react/src/modules/slack/hooks/useSlackChannels.ts
```diff
@@ -46,13 +42,22 @@ export function useSlackChannels(
               knockChannelId: knockSlackChannelId,
             })
             .then((res) => {
-              const { channels } = res;
-              setData({ channels });
+              if (res.code === "ERR_BAD_REQUEST") {
+                const message = res.response?.data?.message
```
connorlindsey, line 46: Should we have a default error message in case there isn't one in the response? Since we're checking on the presence of the error message, I think we should.

packages/react/src/modules/slack/components/ConnectToSlackContainer/styles.css
```diff
@@ -1,31 +1,28 @@
...
-    flex-direction: row;
-    padding: 20px;
-    justify-content: space-between;
-    align-items: flex-start;
-    border-radius: 8px;
-    border: 1px solid var(--Gray-Gray-100, #e4e8ee);
-    background: var(--Gray-White, #fff);
-    box-shadow: 0px 4px 6px -1px rgba(0, 0, 0, 0.1), 0px 2px 4px -1px rgba(0, 0, 0, 0);
-  }
+.rnf-slack-container-container {
```
connorlindsey, line 1: Nice scoping to `rnf-slack-`

packages/client/src/clients/slack/index.ts
```diff
@@ -33,20 +43,40 @@ class SlackClient {
...
+  }: GetSlackChannelsInput) {
+    const result = await this.instance.client().makeRequest({
+      method: "GET",
+      url: `/v1/providers/slack/${knockChannelId}/channels`,
+      params: {
+        access_token_object: {
+          object_id: tenant,
           collection: TENANT_COLLECTION,
         },
         connections_object: {
```
connorlindsey, line 75: Nit: Can probably just pass in `connections_object` instead of reassigning each property

packages/react/src/modules/slack/components/ConnectToSlackButton/ConnectToSlackButton.tsx
```diff
@@ -18,32 +16,70 @@ export const ConnectToSlackButton: React.FC<Props> = ({
...
-      channel_id: knockSlackChannelId,
-      public_key: knock.apiKey,
-      user_token: knock.userToken,
-    }),
-    client_id: slackClientId,
-    scope: "chat:write,chat:write.public,channels:read,groups:read",
-  } as Record<string, string>;
-  const params = new URLSearchParams(rawParams);
+  // Loading states
+  if (connectionStatus === "loading" || connectionStatus === "disconnecting") {
```
connorlindsey, line 28: Nit: would consider moving to separate components but not necessarily something we need to do right now

packages/react/src/modules/slack/hooks/useManageSlackConnection.ts
```diff
@@ -0,0 +1,146 @@
+import { useKnockClient } from "@knocklabs/react-core";
+import { useState, useEffect, useCallback } from "react";
+
+const SLACK_AUTHORIZE_URL = "https://slack.com/oauth/v2/authorize";
+const TENANT_OBJECT_COLLECTION = "$tenants";
```
connorlindsey, line 5: Nit: would consider moving to a constants file if we're using this in multiple places

packages/react/src/modules/slack/hooks/useManageSlackConnection.ts
```diff
@@ -0,0 +1,146 @@
...
+  const rest = errorMessage.substring(1);
+  return firstLetter?.concat(rest).replace("_", " ");
+};
+
+export function useManageSlackConnection(
+  knockSlackChannelId: string,
+  tenant: string,
+  slackClientId: string,
+  redirectUrl?: string,
+): {
```
connorlindsey, line 33: Nit: would move this to a named type

packages/react/src/modules/slack/hooks/useManageSlackConnection.ts
```diff
@@ -0,0 +1,146 @@
...
+          tenant,
+          knockChannelId: knockSlackChannelId,
+        });
+
+        if (authRes.connection?.ok) {
+          setConnectionStatus("connected");
+          return;
+        }
+
+        // This is a normal response for a tenant that doesn't have an access
```
connorlindsey, line 63: Appreciate all the comments explaining the non-obvious slack behavior!

connorlindsey (approved):
This is looking great! A few small comments but don't want to block

packages/react/src/modules/slack/hooks/useSlackChannels.ts
meryldakin: Yes - i'll come back to this one actually though when I spruce up the channel picker, there will be more to do on error handling there anyway

packages/client/src/clients/slack/index.ts
meryldakin: can't annoyingly because of `objectId` -> `object_id`

packages/react/src/modules/slack/hooks/useManageSlackConnection.ts
meryldakin: will do if we reuse these but i might have it scoped to just this file if we can reuse this hook when we introduce another component that uses it

PR #30, opened 2024-01-25
feat: implement automatic socket disconnect/reconnect

connorlindsey:
Watches for document visibility changes and automatically disconnects and reconnects the web socket, after a delay.


packages/client/src/api.ts
```diff
@@ -82,6 +82,18 @@ class ApiClient {
     }
   }
 
+  reconnectSocket() {
+    if (this.socket && !this.socket.isConnected()) {
+      this.socket.connect();
```
connorlindsey, line 87: TIL Phoenix sockets are stateful, so reconnecting handles rejoining channels as well

kylemcd (approved):

packages/client/src/clients/feed/feed.ts
```diff
@@ -667,6 +674,42 @@ class Feed {
       console.warn(`Could not broadcast ${type}, got error: ${e}`);
     }
   }
+
+  /**
+   * Listen for changes to document visibility and automatically disconnect
+   * or reconnect the socket after a delay
+   */
+  private setupAutoSocketManager(
+    auto_manage_socket_connection: boolean | number,
```
cjbell, line 683: Nit: I know we have this as snake cased as the option name, but internally here I would expect it as camelCase

packages/client/src/clients/feed/interfaces.ts
```diff
@@ -22,6 +22,10 @@ export interface FeedClientOptions {
   trigger_data?: GenericData;
   // Optionally enable cross browser feed updates for this feed
   __experimentalCrossBrowserUpdates?: boolean;
+  // Optionally automatically manage socket connections on changes to tab visibility
+  // Defaults to false. If true, defaults to disconnecting after 2000ms when the tab is hidden
+  // Accepts a number as a custom delay duration in milliseconds
+  auto_manage_socket_connection?: boolean | number;
```
cjbell, line 28: 🤔 I wonder if this should be 2 options, one for enabling the behavior and one for setting the disconnect window.

cjbell (commented):
two tiny comments but looks great

packages/client/src/clients/feed/interfaces.ts
connorlindsey: Yeah, was debating that. Having one option could be a little confusing and requires more explanation. Having two options is more verbose and requires some extra logic to handle both cases (not a problem, just something to consider if e.g. a duration is set but the boolean flag is off)

connorlindsey: Going to refactor to have two options. I think it's easier to understand

cjbell (approved):
💯💯💯 beautiful

PR #34, opened 2024-01-30
fix: esm import issues

connorlindsey:
When importing `@knocklabs/react` as esm, `@knocklabs/react-core` gets imported as common js where it should be esm. This updates the build to output `.cjs` files in the cjs build and `.js` in the esm build. Updates package.json to disambiguate the imports.


kylemcd:
Fixed with #48

PR #35, opened 2024-01-31
fix: update popperjs dependency

connorlindsey:
Adds missing popperjs import

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -46,6 +46,7 @@
   "dependencies": {
     "@knocklabs/client": "*",
     "@knocklabs/react-core": "*",
+    "@popperjs/core": "^2.11.8",
     "lodash.debounce": "^4.0.8",
     "react-popper": "^2.3.0",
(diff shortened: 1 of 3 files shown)


packages/react/package.json
```diff
@@ -46,6 +46,7 @@
   "dependencies": {
     "@knocklabs/client": "*",
     "@knocklabs/react-core": "*",
+    "@popperjs/core": "^2.11.8",
```
thomaswhyyou, line 49: Guessing this is a peer dependency of react-popper?

thomaswhyyou (approved):

packages/react/package.json
connorlindsey: Yeah, it's a peer dependency in react-popper, but we also use it for types in `NotificationFeedPopover.tsx`. Not sure why it being missing didn't break the build. Issue only surfaced when there was a conflict with a user's app which also had react-popper installed.

PR #37, opened 2024-01-31
chore: make configs publishable for use in other repos

kylemcd:
Making the eslint-config and typescript-config publishable so that I can use the configs in telegraph.


connorlindsey (changes requested):
Looks good! Will also need a changeset. You can add one by running `yarn changeset` from the root of the monorepo and going through the cli. The changeset should just affect these two packages

connorlindsey (approved):

PR #39, opened 2024-01-31
feat(kno-5138): enable popup window for slack auth flow

meryldakin:
Adds a pop up window for the slack auth flow rather than using a redirect for this. Relies on switchboard PR to return the window close html: https://github.com/knocklabs/switchboard/pull/1909

### DIFF
--- a/packages/react/src/modules/slack/components/ConnectToSlackButton/ConnectToSlackButton.tsx
+++ b/packages/react/src/modules/slack/components/ConnectToSlackButton/ConnectToSlackButton.tsx
@@ -2,6 +2,8 @@ import { SlackIcon } from "../SlackIcon";
 import "./styles.css";
 
 import { useManageSlackConnection } from "../../hooks/useManageSlackConnection";
+import { useEffect } from "react";
+import { useKnockClient } from "@knocklabs/react-core";
 
 type Props = {
   tenant: string;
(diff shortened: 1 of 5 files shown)


packages/react/src/modules/slack/components/ConnectToSlackButton/ConnectToSlackButton.tsx
```diff
@@ -10,19 +12,74 @@ type Props = {
...
+
+  const left = innerWidth / 2 - width / 2 + screenLeft;
+  const top = innerHeight / 2 - height / 2 + screenTop;
+
+  // Window features
+  const features = `width=${width},height=${height},top=${top},left=${left}`;
+
+  window.open(url, "Slack OAuth", features);
+};
+
```
kylemcd, line 37: This is so cool! Didn't know something like that was possible.

packages/react/src/modules/slack/components/ConnectToSlackButton/ConnectToSlackButton.tsx
```diff
@@ -42,7 +99,7 @@ export const ConnectToSlackButton: React.FC<Props> = ({
   if (connectionStatus === "error") {
     return (
       <a
-        href={buildSlackAuthUrl()}
+        onClick={() => openSlackOauthPopup(buildSlackAuthUrl())}
         className="rnf-slackConnect-button rnf-slackConnect-button--error"
         onMouseEnter={() => setActionLabel("Reconnect")}
         onMouseLeave={() => setActionLabel("")}
```
kylemcd, line 105: Semantically this should be a button now that you're using an onClick instead of an href

packages/react/src/modules/slack/components/ConnectToSlackButton/ConnectToSlackButton.tsx
```diff
@@ -59,7 +116,7 @@ export const ConnectToSlackButton: React.FC<Props> = ({
   if (connectionStatus === "disconnected") {
     return (
       <a
-        href={buildSlackAuthUrl()}
+        onClick={() => openSlackOauthPopup(buildSlackAuthUrl())}
         className="rnf-slackConnect-button rnf-slackConnect-button--disconnected"
       >
```
kylemcd, line 121: Same here

kylemcd (changes requested):

connorlindsey (approved):
Agreed on switching from `<a>` to `<button>`. Otherwise this looks great!

packages/react/src/modules/slack/components/ConnectToSlackButton/ConnectToSlackButton.tsx
meryldakin: i didn't either before this 🙃

kylemcd (approved):
Nice!

PR #40, opened 2024-01-31
fix: eslint-config devDeps not accessible, moving to deps

kylemcd:
When installing `@knocklabs/eslint-config` because the dependencies are set to `devDependencies`, they're not installed along with the config. This means each of those dependencies would need to be installed on their own. 

This moves these `devDependencies` to `dependencies` to make install smoother.

### DIFF
--- a/.changeset/popular-days-worry.md
+++ b/.changeset/popular-days-worry.md
@@ -1,0 +1,5 @@
+---
+"@knocklabs/eslint-config": patch
+---
+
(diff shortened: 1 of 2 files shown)


kylemcd:
> Looks good, but would run an install and push again in case there are conflicts with the lock file during deploy (I don't think it'll be a problem but just to be safe)

Yeah I regenerated the lock file but it was the same as before. No changes in git.

PR #41, opened 2024-01-31
chore: add testing set up and base coverage for React SDKs

connorlindsey:
Sets up Vitest and testing-library for react and react-core with a few base test cases. 

Holding off on react native as Vitest doesn't support it. There are some workarounds we could consider like rendering using react-native-web during testing or using Jest. https://github.com/vitest-dev/vitest/issues/1321


kylemcd (approved):
Looks awesome! test-utils helper is great too!

PR #45, opened 2024-02-02
fix: strip socket manager options from http requests

connorlindsey:
Updates the README and example app to show using the automatic socket manager. Also removes the added options from the request query params.

### DIFF
--- a/packages/react/README.md
+++ b/packages/react/README.md
@@ -6,7 +6,7 @@ A set of components for integrating [Knock](https://knock.app) into a React appl
 
 [See a live demo](https://knock-in-app-notifications-react.vercel.app/)
 
-![In-app feed component example](NotificationFeed.png)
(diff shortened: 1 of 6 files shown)


packages/react/README.md
```diff
@@ -6,7 +6,7 @@ A set of components for integrating [Knock](https://knock.app) into a React appl
 
 [See a live demo](https://knock-in-app-notifications-react.vercel.app/)
 
-![In-app feed component example](NotificationFeed.png)
+![In-app feed component example](https://github.com/knocklabs/javascript/raw/main/packages/react/NotificationFeed.png)
```
connorlindsey, line 9: The image doesn't load on the NPM package page so I think linking to the actual image instead of relative path should fix it.

kylemcd (approved):

PR #46, opened 2024-02-02
feat: switch to date-fns intlFormatDistance to reduce bundle size

connorlindsey:
Switches from `formatDistance` to `intlFormatDistance` which uses the built-in browser `Intl` object for formatting relative times: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat/RelativeTimeFormat#browser_compatibility

This should greatly reduce the bundle size. 

TODO:
- [x] Update relevant docs / readmes
- [x] Double check package size. I built it locally and ran `npm pack` but want to double check the difference


packages/react-core/src/modules/core/utils.ts
```diff
@@ -1,13 +1,13 @@
...
+import { parseISO, intlFormatDistance } from "date-fns";
 import { ReactNode } from "react";
 
 export function formatBadgeCount(count: number): string | number {
   return count > 9 ? "9+" : count;
 }
 
 type FormatTimestampOptions = {
-  locale?: Locale;
+  locale?: string | string[];
```
kylemcd, line 10: Does it make sense to strongly type this to available Locales? Not sure what that involves but feels like that would be a good DX?

kylemcd (approved):

packages/react-core/src/modules/core/utils.ts
connorlindsey: Agreed it'd be a nice DX improvement but I didn't find a good canonical list e.g. a few of the packages I saw had a lot of helper methods which would bloat the bundle size again. Unfortunately the Intl module doesn't have an out of the box method for this (they also type locale as `string | string[]`. From what I understand, it can depend on the browser implementation as well. Given that, I think we should punt on that.

cjbell (approved):
this is awesome 💯

PR #48, opened 2024-02-07
fix: build process causing esm issues

kylemcd:


packages/react/vite.config.ts
```diff
@@ -61,8 +69,7 @@ export default defineConfig(({ mode }) => {
 
                 if (file?.type === "chunk") {
                   // Replace .css imports and requires
-                  const pattern =
```
connorlindsey, line 64: We need to keep the `require` portion of the regex as that strips the unnecessary `.css.js` imports from the cjs build. For example, if you build and look at `react/dist/cjs/modules/core/components/Button/Button.js` you'll see it includes `require("./styles.css.js");`

`/(import\s+['"].+\.css\..*?['"];?)|(require\(['"][^()]+\.css(\.js)?['"]\);?)/g`

packages/react/vite.config.ts
```diff
@@ -9,34 +13,35 @@ import execute from "rollup-plugin-execute";
...
         outDir: "dist/types",
       }),
       react(),
-      noBundlePlugin({ root: "./src" }),
+      // splitVendorChunkPlugin(),
+      noBundlePlugin({ root: resolve(__dirname, "src") }),
     ],
     build: {
-      outDir: CJS ? "dist/cjs" : "dist/esm",
+      // cssMinify: "lightningcss",
```
connorlindsey, line 28: Do we want to change css minifier? I didn't look too closely at the different options.

packages/react/vite.config.ts
```diff
@@ -50,7 +55,10 @@ export default defineConfig(({ mode }) => {
             // Move index.css to root of dist
             `mv dist/esm/index.css dist/index.css`,
             // Delete extra .css.js files
-            `rm dist/**/*.css.*`,
+            `find ./dist -name "*.css.js" -delete`,
```
connorlindsey, line 58: Nit: I think we could get this to a single line catching anything `.css.*`: ``find ./dist -name "*.css.*" -delete``

connorlindsey (commented):
Tested this against Next 12/13/14, Blitz, and CRA and I think this is a good fix for the ESM issues we were seeing, just a bit of clean up around comments/imports

packages/react/vite.config.ts
kylemcd: Ah I was testing something out, isn't a big deal here, will just remove the commented code.

packages/react/package.json
```diff
@@ -68,6 +69,7 @@
     "eslint-plugin-react-hooks": "^4.6.0",
     "eslint-plugin-react-refresh": "^0.4.4",
     "jsdom": "^24.0.0",
+    "lightningcss": "^1.23.0",
```
connorlindsey, line 72: If we're not using lightningcss, let's remove the dependency and re-install to update the lock file

packages/react/vite.config.ts
```diff
@@ -9,34 +9,33 @@ import execute from "rollup-plugin-execute";
...
-        // External packages that should not be bundled
-        external: ["react", "react-dom"],
+        external: ["react"],
         output: {
           interop: "compat",
-          format: formats[0],
           globals: {
             react: "React",
           },
+          // External packages that should not be bundled
```
connorlindsey, line 38: Nit: Comment should be above the `external` option

connorlindsey (approved):
Little bit more clean up but approving to not block getting this out

Issue #23, opened 2023-07-14
Knock stops receiving notifications if JWT changes


Diggsey:
When using authentication, if the JWT provided to `<KnockFeedProvider>` changes, then notifications stop being received in the inner `<NotificationFeedPopover>`.

This occurs only in chrome, and seems to be related to the WebSocket connection being closed.

connorlindsey:
Hi @Diggsey, we just released a new version of our JS client and React components which resolves this issue. Please upgrade to the latest versions (https://www.npmjs.com/package/@knocklabs/react/v/0.1.8) and let us know if you see any other issues. Thanks!

PR #49, opened 2024-02-08
feat: add token expiration handler

connorlindsey:
## Description
This adds the option to pass a callback to the Knock client when authenticating a user that will be called when a token expires, making it easier to refresh the user's token.


packages/react-core/src/modules/core/hooks/useAuthenticatedKnockClient.ts
```diff
@@ -1,23 +1,24 @@
...
-import Knock, { KnockOptions } from "@knocklabs/client";
+import Knock, { AuthenticateOptions, KnockOptions } from "@knocklabs/client";
 
 function useAuthenticatedKnockClient(
   apiKey: string,
   userId: string,
-  userToken: string | undefined,
+  userToken?: string,
   options: KnockOptions = {},
+  authenticateOptions?: AuthenticateOptions,
```
cjbell, line 9: Can we merge these two options, please?

packages/client/src/knock.ts
```diff
@@ -74,11 +103,46 @@ class Knock {
...
+      // | ----------------- | ----------------------- |
+      // ^ now               ^ expiration offset       ^ expires at
+      const msInFuture = expiresAtMs - timeBeforeExpirationInMs - nowMs;
+
+      this.tokenExpirationTimer = setTimeout(async () => {
+        const newToken = await callbackFn(this.userToken as string, decoded);
+        this.userToken = newToken;
+
+        // Resync socket connection
+        this.createApiClient();
```
cjbell, line 137: Do we need to teardown old connection before starting a new one here?

packages/client/src/knock.ts
```diff
@@ -74,11 +103,46 @@ class Knock {
...
+    }
     if (!this.apiClient) return;
     if (this.apiClient.socket) {
       this.apiClient.socket.disconnect();
     }
   }
+
+  private async maybeScheduleUserTokenExpiration(
+    callbackFn: UserTokenExpiringCallback,
+    timeBeforeExpirationInMs: number = 1000,
```
cjbell, line 117: I think we should set this to be 30_000 as a default to give plenty of time to refresh

examples/nextjs-example/pages/index.tsx
```diff
@@ -51,12 +51,26 @@ export default function Home() {
...
       host={process.env.NEXT_PUBLIC_KNOCK_HOST}
+      authenticateOptions={{
+        onUserTokenExpiring: async () => {
+          // Refresh the user token 1s before it expires
+          const res = await fetch(`/api/auth?id=${userId}`);
+          const json = await res.json();
+          return json.userToken;
+        },
+        timeBeforeExpirationInMs: 500,
+      }}
```
cjbell, line 65: The only thing I was thinking if we should make these top level props instead of under an object. Might be cleaner?

cjbell (commented):
few initial comments, but this is awesome!

packages/client/src/knock.ts
connorlindsey: Good question. I initially thought no since there was already a teardown/deconstructor for the api client, but yeah, maybe we should disconnect the socket to keep everything fresh.

connorlindsey: This does make this issue more apparent: https://github.com/knocklabs/javascript/issues/23

Toasts come through but the notification count doesn't update after the token changes

cjbell: Does this disconnect/reconnect solve the real-time issue? Do we need to do anything at the feed layer to reconnect the connected channels or does instantiating a new API client just handle that?

connorlindsey: It doesn't. Starting to investigate more and I think it might be an issue with how we memoize the knock and feed clients.

cjbell (approved):

PR #53, opened 2024-02-12
feat(kno-5213): accept pagination params and split out connected channels from all slack channels hooks

meryldakin:
This PR allows us to paginate the slack channels query by introducing ReactQuery (note: the provider will be moved under the upcoming SlackProvider when it's available). 

It also splits up our hooks for getting all of the available slack channels and only getting the connected channels because now that we have pagination, it makes it a lot easier for us to get and set that connected channel data when it's split from the whole list of slack channels.

There's other minor refactoring throughout, but a lot will still change when we introduce a shared context for the connected state of all the slack components, just to keep in mind as you're reviewing.


packages/react/src/modules/slack/hooks/useManageSlackConnection.ts
```diff
@@ -123,7 +128,7 @@ export function useManageSlackConnection(
         user_token: knock.userToken,
       }),
       client_id: slackClientId,
-      scope: SLACK_SCOPES,
+      scope: DEFAULT_SLACK_SCOPES.join(","),
```
meryldakin, line 131: will be introducing scopes as a param so made this small change here

packages/client/src/clients/objects/index.ts
```diff
@@ -0,0 +1,63 @@
...
+
+type SetChannelDataInput = {
+  objectId: string;
+  collection: string;
+  channelId: string;
+  data: any;
+  userId: string;
+};
+
+class ObjectClient {
```
meryldakin, line 19: Added this client so that we could move these generic calls out of the slack client, which previously had the `setConnectedChannels` function that essentially was just setting channel data.

packages/react-core/src/modules/core/context/KnockProvider.tsx
```diff
@@ -45,7 +48,15 @@ export const KnockProvider: React.FC<KnockProviderProps> = ({
         knock,
       }}
     >
-      <KnockI18nProvider i18n={i18n}>{children}</KnockI18nProvider>
+      <KnockI18nProvider i18n={i18n}>
+        {/**
+         * TODO 5211: Move this QueryClientProvider into the SlackProvider when we create it MKD
```
meryldakin, line 53: note on moving this once the slack provider is introduced in the next PR

packages/react/src/App.tsx
```diff
@@ -1,8 +1,6 @@
 import {
   Button,
   ButtonGroup,
-  ConnectToSlackButton,
```
meryldakin, line 4: unused

packages/react/src/modules/slack/components/ConnectedSlackChannelPicker/ConnectedSlackChannelPicker.tsx
```diff
@@ -15,85 +18,67 @@ export const ConnectedSlackChannelPicker: React.FC<Props> = ({
...
+  });
 
-  const isLoading = isSettingChannels || isLoadingSlackChannels;
-
-  const { channels } = data;
+  const { data: connectedChannels, updateConnectedChannels } =
+    useConnectedChannels({
+      connectionsObject,
+      knockSlackChannelId,
+    });
```
meryldakin, line 36: split off this into its own hook

packages/react/src/modules/slack/components/ConnectedSlackChannelPicker/ConnectedSlackChannelPicker.tsx
```diff
@@ -111,39 +96,22 @@ export const ConnectedSlackChannelPicker: React.FC<Props> = ({
             key={"select"}
             className="rnf-header"
             onClick={() => {
-              if (!errorMessage) {
```
meryldakin, line 114: will be introducing errors back when we have a context for the connected state of these components, doesn't make sense to deal with it in this PR

packages/react/src/modules/slack/components/ConnectedSlackChannelPicker/ConnectedSlackChannelPicker.tsx
```diff
@@ -152,41 +120,43 @@ export const ConnectedSlackChannelPicker: React.FC<Props> = ({
...
-                        fill="#9EA0AA"
-                      />
-                    </svg>
-                  </span>
-                </div>
+              {slackChannels?.map((slackChannel: SlackChannel) => (
+                <SlackChannelOption
```
meryldakin, line 149: moved into its own component so we could more easily handle the `isConnected` state

packages/react/src/modules/slack/hooks/useConnectedChannels.ts
```diff
@@ -0,0 +1,74 @@
...
+};
+
+type UseSlackChannelOutput = {
+  data: SlackChannelConnection[] | null;
+  updateConnectedChannels: (
+    connectedChannels: SlackChannelConnection[],
+  ) => void;
+};
+
+export function useConnectedChannels({
```
meryldakin, line 19: new hook that's using the `getChannelData` function and endpoint in Switchboard

packages/react/src/modules/slack/hooks/useConnectedChannels.ts
```diff
@@ -0,0 +1,74 @@
...
+      await knockClient.objects.setChannelData({
+        objectId,
+        collection,
+        channelId: knockSlackChannelId,
+        data: { connections: channelsToSendToKnock },
+        userId: knockClient.userId!,
+      });
+      setShouldRefetch(true);
+    } catch (error) {
+      console.log(error);
```
meryldakin, line 66: this will add to an error state when we put that back in

packages/react/src/modules/slack/hooks/useSlackChannels.ts
```diff
@@ -1,77 +1,87 @@
...
-              }
+  const fetchChannels = ({ pageParam }: { pageParam: string }) => {
+    return knockClient.slack.getChannels({
+      tenant,
+      connectionsObject,
+      knockChannelId: knockSlackChannelId,
+      queryOptions: {
+        ...queryOptions,
+        cursor: pageParam,
+        limit: queryOptions?.limitPerPage || LIMIT_PER_PAGE,
```
meryldakin, line 59: this now includes the appropriate query options and removes the connections object, which we don't need anymore here

packages/react/src/modules/slack/hooks/useSlackChannels.ts
```diff
@@ -1,77 +1,87 @@
...
+  };
+
+  const { data, isLoading, isFetching, fetchNextPage, hasNextPage, refetch } =
+    useInfiniteQuery({
+      queryKey: ["slackChannels"],
+      queryFn: fetchChannels,
+      initialPageParam: "",
+      getNextPageParam: (lastPage) =>
+        lastPage?.next_cursor === "" ? null : lastPage?.next_cursor,
+    });
```
meryldakin, line 71: this is an "infiniteQuery" in that it will keep all of the fetched data in state instead of "paging" through and erasing the previous fetch from state on each call.

packages/react/src/modules/slack/hooks/useSlackChannels.ts
```diff
@@ -1,77 +1,87 @@
...
+      ?.flatMap((page) => page?.slack_channels)
+      .filter((channel) => !!channel) || [];
 
-      setIsLoading(true);
-      fetchChannels();
+  const maxCount = queryOptions?.maxCount || MAX_COUNT;
+
+  useEffect(() => {
+    if (hasNextPage && !isFetching && slackChannels?.length < maxCount) {
+      fetchNextPage();
```
meryldakin, line 82: this makes sure to continue making the query until we reach the maximum amount of requested channels

packages/react/package.json
```diff
@@ -53,6 +53,8 @@
     "@knocklabs/client": "*",
     "@knocklabs/react-core": "*",
     "@popperjs/core": "^2.11.8",
+    "@tanstack/react-query": "^5.18.1",
+    "@tanstack/react-query-devtools": "^5.20.1",
```
connorlindsey, line 57: Nit: Let's make this a dev dependency

packages/react/src/modules/slack/components/ConnectedSlackChannelPicker/ConnectedSlackChannelPicker.tsx
```diff
@@ -15,85 +18,67 @@ export const ConnectedSlackChannelPicker: React.FC<Props> = ({
...
-    }
-
     if (isLoading) {
       return "Fetching channels...";
     }
 
-    const numberConnectedChannels = channels.filter(
-      (channel: Channel) => channel.connected,
-    ).length;
+    if (!isLoading && slackChannels.length === 0) {
```
connorlindsey, line 66: Nit: !isLoading check is unnecessary because of the previous isLoading check

packages/react/src/modules/slack/hooks/useSlackChannels.ts
```diff
@@ -1,77 +1,86 @@
 import { useKnockClient } from "@knocklabs/react-core";
```
connorlindsey, line 1: Would it make sense to move any of these hooks to react-core? The components that are currently using them would only work for react web but the underlying data fetching and react query should work out of the box with React Native.

connorlindsey (approved):
This looks awesome! I think React Query is a great choice here and makes the refetching logic pretty clean ✨

packages/react/src/modules/slack/hooks/useSlackChannels.ts
meryldakin: yeah that's totally right. Let me do that in the next PR where I'm introducing the provider.

PR #55, opened 2024-02-12
add slack provider

meryldakin:


packages/react-core/src/modules/slack/hooks/useSlackConnectionStatus.ts
```diff
@@ -97,57 +87,14 @@ export function useManageSlackConnection(
     checkAuthStatus();
   }, [connectionStatus, tenant, knockSlackChannelId, knock.slack]);
 
-  const disconnectFromSlack = useCallback(async () => {
```
meryldakin, line 100: moved to `useManageSlackCOnnection` hook

packages/react-core/src/modules/slack/hooks/useManageSlackConnection.ts
```diff
@@ -0,0 +1,87 @@
...
+  "groups:read",
+  "groups:write",
+];
+
+type UseManageSlackConnectionOutput = {
+  buildSlackAuthUrl: () => string;
+  disconnectFromSlack: () => void;
+};
+
+function useManageSlackConnection(
```
meryldakin, line 19: this was just moved and I removed the connection data from it that's now in the useSlackConnectionStatus hook

packages/react-core/src/modules/slack/context/KnockSlackProvider.tsx
```diff
@@ -0,0 +1,83 @@
...
+import * as React from "react";
+
+import { slackProviderKey } from "../../core";
+import { useKnockClient } from "../../core";
+import { ConnectionStatus } from "../hooks/useSlackConnectionStatus";
+
+const queryClient = new QueryClient();
+
+export interface KnockSlackProviderState {
+  knock: Knock;
```
cjbell, line 13: I don't think we should reexpose the knock instance, better to just have anything that needs it call `useKnockClient`

packages/react-core/src/modules/slack/context/KnockSlackProvider.tsx
```diff
@@ -0,0 +1,83 @@
...
+    </SlackProviderStateContext.Provider>
+  );
+};
+
+export const useKnockSlackClient = (): KnockSlackProviderState => {
+  const context = React.useContext(
+    SlackProviderStateContext,
+  ) as KnockSlackProviderState;
+  if (context === undefined) {
+    throw new Error("useKnock must be used within a KnockProvider");
```
cjbell, line 80: 👀 this error needs updating

packages/react-core/src/modules/slack/hooks/useManageSlackConnection.ts
```diff
@@ -0,0 +1,87 @@
+import { useKnockSlackClient } from "..";
+import { useCallback } from "react";
+
+const SLACK_AUTHORIZE_URL = "https://slack.com/oauth/v2/authorize";
+const TENANT_OBJECT_COLLECTION = "$tenants";
```
cjbell, line 5: it might be nice to expose this under the client somewhere as a constant

packages/react-core/src/modules/slack/hooks/useManageSlackConnection.ts
```diff
@@ -0,0 +1,87 @@
...
+  disconnectFromSlack: () => void;
+};
+
+function useManageSlackConnection(
+  slackClientId: string,
+  redirectUrl?: string,
+): UseManageSlackConnectionOutput {
+  const {
+    setConnectionStatus,
+    knock,
```
cjbell, line 25: This can be `const {knock} = useKnockClient()`

cjbell (commented):
this looks awesome!!

packages/react-core/src/modules/slack/context/KnockSlackProvider.tsx
meryldakin: ok cool

examples/slack-connect-example/pages/index.tsx
```diff
@@ -39,124 +41,104 @@ export default function Home() {
...
-        }}
+    <KnockProvider
+      apiKey={process.env.NEXT_PUBLIC_KNOCK_CLIENT_ID!}
+      userId={user.id}
+      host={process.env.NEXT_PUBLIC_KNOCK_API_URL}
+      userToken={localStorage.getItem("knock-user-token")!}
+    >
+      <KnockSlackProvider
+        knockSlackChannelId={process.env.NEXT_PUBLIC_KNOCK_SLACK_CHANNEL_ID!}
+        tenant={tenant}
```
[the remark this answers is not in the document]
meryldakin: I think that we may not be shipping this example app as is because jeff is working on a separate one to showcase this work - if we do end up keeping this one I can throw if they don't exist

kylemcd (approved):

PR #56, opened 2024-02-13
chore: upgrade to yarn modern and update local package references

connorlindsey:
## Description
Updates the repo to use Yarn 4.x so that we can use `workspace:^` to manage cross-package dependencies. This should fix any issues caused by using `"*"` for our version requirements and packages getting out of sync when installed by end users (e.g. `@knocklabs/react` gets updated but is still using an older version of `@knocklabs/react-core`)


kylemcd (commented):
Nice! I added these to the .gitignore in telegraph:

```
# Yarn Berry
.yarn/unplugged
.yarn/install-state.gz
.yarn/build-state.yml
.pnp*
```

PR #62, opened 2024-02-15
feat(kno-5241): if max channels reached, use text input to add channel

meryldakin:
This is a v1 text input for adding channels that I'd like to get in for us to play with and see what it needs. 
Basically we want to restrict using the combobox to only work if you have fewer than 1000 slack channels. if you have more than that, we'll intentionally degrade the experience to help with performance, which is where we have this input box come in.

It's very stupid, it just takes a text string and posts it to the setChannelData endpoint as an entry in channel data. It does not check if it belongs to a real slack channel (and it shouldn't this is out of scope). It does check if the channel id already exists in the connected channel ids. Once it's in, we can tweak the design.
https://www.loom.com/share/7fda538f23fc4fd789bd837f2eb3cc44

### DIFF
--- a/packages/react/src/modules/slack/components/SlackChannelCombobox/AddConnectedSlackChannelInput.tsx
+++ b/packages/react/src/modules/slack/components/SlackChannelCombobox/AddConnectedSlackChannelInput.tsx
@@ -1,0 +1,72 @@
+import { SlackChannelConnection } from "@knocklabs/client";
+import { useState } from "react";
+
+import { SlackIcon } from "../SlackIcon";
+
+import ConnectionErrorInfoBoxes from "./ConnectionErrorInfoBoxes";
+import "./styles.css";
+
+const AddConnectedSlackChannelInput = ({
+  isErrorState,
+  connectedChannels = [],
+  updateConnectedChannels,
+  connectedChannelsError,
+  connectedChannelsUpdating,
+}: {
+  isErrorState: boolean;
+  connectedChannels: SlackChannelConnection[];
+  updateConnectedChannels: (channels: SlackChannelConnection[]) => void;
+  connectedChannelsError: string | null;
+  connectedChannelsUpdating: boolean;
+}) => {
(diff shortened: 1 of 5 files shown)


packages/react/src/modules/slack/components/SlackChannelCombobox/ConnectionErrorInfoBoxes.tsx
```diff
@@ -0,0 +1,38 @@
+import { useKnockSlackClient } from "@knocklabs/react-core";
+
+import InfoIcon from "./icons/InfoIcon";
+
+const ConnectionErrorInfoBoxes = () => {
+  const { connectionStatus } = useKnockSlackClient();
+  return (
+    <>
+      {connectionStatus === "disconnected" && (
```
connorlindsey, line 9: Looks like the only difference is the message, so I think these could be combined to:
```
        <div className="rnf-disconnected-info-container">
          <span>
            <InfoIcon />
          </span>

          <div className="rnf-info-container-text">
           {<connection message>}
          </div>
        </div>
 ```
We could either have a conditional in the jsx or a separate function that returns the message.

File: packages/react/src/modules/slack/components/SlackChannelCombobox/AddConnectedSlackChannelInput.tsx [hunk 2]
```diff
@@ -0,0 +1,72 @@
...
+      return setLocalError(`"${value}" already connected.`);
+    }
+
+    const channelsToSendToKnock = [...connectedChannels, { channel_id: value }];
+    updateConnectedChannels(channelsToSendToKnock);
+    setValue("");
+  };
+
+  return (
+    <div style={{ display: "flex", gap: "4px", alignItems: "center" }}>
```
connorlindsey, line 49: Nit: Should clean these inline styles up (but can address it once the design is finalized)

connorlindsey (approved):
Couple minor style comments, but looks good. Will approve to not block

PR #65, opened 2024-02-16
chore(kno-5291): rename all slack components

meryldakin:
Uses the updated names for all components and hooks.


kylemcd (approved):

packages/react/src/modules/slack/components/SlackChannelCombobox/SlackChannelCombobox.tsx
JEverhart383, line 1: Only thing I can see is this prop: https://www.notion.so/knockapp/Component-API-design-7775383f38004b29af41ae5f7cec2da0?pvs=4#9506493635dd4552b5ebce6eee6015c6
`slackChannelsRecipientObject` but maybe that is too verbose given the other renaming

JEverhart383 (approved):
Just one comment about one of the props that we had a back and forth on, but with some of the other renaming I could go either way now, so totally non-blocking.

packages/react/src/modules/slack/components/SlackChannelCombobox/SlackChannelCombobox.tsx
meryldakin: just missed this one! will rename

PR #66, opened 2024-02-16
fix: gracefully handle credential changes

cjbell:
This is quite a few updates, but it essentially boils down to:

- Gracefully handling userId and userToken changes by reauthenticating and reconnecting real-time connections
- Memoizing options to prevent the API client from being recreated when unnecessary
- Removing `store.destroy()` when tearing down a feed instance which was causing the state management to get disconnected

### DIFF
--- a/examples/nextjs-example/pages/index.tsx
+++ b/examples/nextjs-example/pages/index.tsx
@@ -13,7 +13,7 @@ import {
   KnockProvider,
   NotificationFeedContainer,
 } from "@knocklabs/react";
-import { useCallback, useEffect, useState } from "react";
+import { useCallback, useState } from "react";
 import { IoDocument, IoLogoGithub } from "react-icons/io5";
 
 import NotificationFeed from "../components/NotificationFeed";
@@ -33,26 +33,17 @@ const TenantLabels = {
 
 export default function Home() {
   const { userId, isLoading, userToken } = useIdentify();
-  const [currentUserToken, setCurrentUserToken] = useState(null);
   const [tenant, setTenant] = useState(Tenants.TeamA);
-
-  useEffect(() => {
-    if (userToken) {
-      setCurrentUserToken(userToken);
-    }
-  }, [userToken]);
 
   const tokenRefreshHandler = useCallback(async () => {
     // Refresh the user token 1s before it expires
     const res = await fetch(`/api/auth?id=${userId}`);
     const json = await res.json();
 
-    setCurrentUserToken(json.userToken);
+    return json.userToken;
+  }, [userId]);
 
-    return;
-  }, [userId, setCurrentUserToken]);
-
-  if (isLoading || !currentUserToken) {
+  if (isLoading || !userToken) {
     return (
       <Flex
         alignItems="center"
@@ -68,7 +59,7 @@ export default function Home() {
   return (
     <KnockProvider
       userId={userId}
-      userToken={currentUserToken || undefined}
+      userToken={userToken}
       apiKey={process.env.NEXT_PUBLIC_KNOCK_PUBLIC_API_KEY!}
       host={process.env.NEXT_PUBLIC_KNOCK_HOST}
       onUserTokenExpiring={tokenRefreshHandler}
--- a/examples/nextjs-example/pages/index.tsx
+++ b/examples/nextjs-example/pages/index.tsx
@@ -13,7 +13,7 @@ import {
   KnockProvider,
   NotificationFeedContainer,
 } from "@knocklabs/react";
-import { useState } from "react";
+import { useCallback, useEffect, useState } from "react";
 import { IoDocument, IoLogoGithub } from "react-icons/io5";
 
 import NotificationFeed from "../components/NotificationFeed";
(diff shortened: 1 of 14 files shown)


packages/react-core/src/modules/core/context/KnockProvider.tsx
```diff
@@ -27,23 +27,40 @@ export interface KnockProviderProps {
 
   // i18n translations
   i18n?: I18nContent;
+
+  logLevel?: LogLevel;
```
cjbell, line 31: Added this so folks can set a log level

packages/react-core/src/modules/core/hooks/useAuthenticatedKnockClient.ts
```diff
@@ -10,13 +22,25 @@ function useAuthenticatedKnockClient(
...
+
+    // If the userId and the userToken changes then just reauth
+    if (
+      currentKnock &&
+      currentKnock.isAuthenticated() &&
+      (currentKnock.userId !== userId || currentKnock.userToken !== userToken)
+    ) {
+      authenticateWithOptions(currentKnock, userId, userToken, options);
+      return currentKnock;
+    }
```
cjbell, line 35: This allows the client to reauthenticate when the user Id or userToken changes, without needing to reinit the entire Knock client

packages/react-core/src/modules/feed/context/KnockFeedProvider.tsx
```diff
@@ -1,20 +1,20 @@
...
-import useNotifications from "../hooks/useNotifications";
 import { feedProviderKey } from "../../core/utils";
-import { useKnockClient } from "../../core";
+import useNotifications from "../hooks/useNotifications";
 
 export interface KnockFeedProviderState {
   knock: Knock;
   feedClient: Feed;
-  useFeedStore: UseStore<FeedStoreState>;
+  useFeedStore: UseBoundStore<FeedStoreState>;
```
cjbell, line 17: Cleaning up deprecated zustand

packages/react-core/src/modules/feed/hooks/useNotifications.ts
```diff
@@ -3,23 +3,25 @@ import { useMemo, useRef } from "react";
...
-  feedId: string,
+  feedChannelId: string,
   options: FeedClientOptions = {},
 ) {
   const feedClientRef = useRef<Feed | null>();
 
   return useMemo(() => {
-    if (feedClientRef.current) feedClientRef.current.teardown();
+    if (feedClientRef.current) {
+      feedClientRef.current.dispose();
```
cjbell, line 13: Changed this to dispose so we destroy the instance when it's no longer needed

examples/nextjs-example/pages/index.tsx
```diff
@@ -33,9 +33,26 @@ const TenantLabels = {
...
+
+  const tokenRefreshHandler = useCallback(async () => {
+    // Refresh the user token 1s before it expires
+    const res = await fetch(`/api/auth?id=${userId}`);
+    const json = await res.json();
+
+    setCurrentUserToken(json.userToken);
+
+    return;
+  }, [userId, setCurrentUserToken]);
```
kylemcd, line 53: Could you explain a little more why we need `currentUserToken`? Would it make more sense for the example to have `setUserToken` come from `useIdentify` instead? 

I also might just be missing something here? If that's the case it might make sense to write a comment explaining what's going on to the engineer viewing this example.

packages/client/src/api.ts
```diff
@@ -1,7 +1,7 @@
 import axios, { AxiosInstance, AxiosRequestConfig } from "axios";
+import { AxiosError } from "axios";
```
kylemcd, line 2:
```suggestion
import axios, type { AxiosError, AxiosInstance, AxiosRequestConfig } from "axios";
```

examples/nextjs-example/pages/index.tsx
cjbell: Oh yeah, that's a great point. Let me clean this up actually, I was just trying out a different way to manage setting this userToken so on refresh the value was consistent .

kylemcd (approved):

examples/nextjs-example/pages/index.tsx
kylemcd: Nice!

packages/react-core/src/modules/core/context/KnockProvider.tsx
```diff
@@ -27,23 +27,40 @@ export interface KnockProviderProps {
...
   timeBeforeExpirationInMs,
   children,
   i18n,
 }) => {
-  const knock = useAuthenticatedKnockClient(apiKey, userId, userToken, {
-    host,
-    onUserTokenExpiring,
-    timeBeforeExpirationInMs,
-  });
+  // We memoize the options here so that we don't create a new object on every re-render
```
connorlindsey, line 45: Do we want to address this now or in a follow up pr?

connorlindsey (approved):
This'll need a changeset targeting the client and react packages so it triggers a new version. Otherwise, looks really good!

PR #67, opened 2024-02-20
fix(kno-5254): hide inaccessible connected channels

meryldakin:
This handles the case where a user has a slack channel connected in Knock, but it isn't found in the list of Slack channels from the API, potentially because the bot doesn't have access to it (like a private channel it doesn't belong to). In this case it also wouldn't be able to post to the channel so it is accurate to show to the user that it is not connected. 

It also updates some minor styling issues.


meryldakin:
see the real PR here: https://github.com/knocklabs/javascript/pull/69

PR #75, opened 2024-02-22
fix(kno-5266): slackkit: css updates + changes from QA

meryldakin:
This handles several changes found necessary during QA. The bulk of it is a refactor of the stylesheets so that everything is standardized and named properly.

These are the other items handled:
*Design*
- [ ]  Align the icon within the searchbar
- [ ]  Clear the search text if you click outside of the combobox
- [ ]  Align search icon, channel icon, and text within the slack options
- [ ]  Add color fade to the hover state

*Combobox functionality*
- [ ]  Make sure you can click anywhere outside of the combobox to exit (see Jeff’s loom)
- [ ]  When you click the combobox, you should immediately be able to type (search input should be focused)


packages/react/src/modules/slack/components/SlackAddChannelInput/styles.css
```diff
@@ -0,0 +1,32 @@
+.rsk-connect-channel {
```
cjbell, line 1: Nit: spacing in this file is off, looks like tabs or 4 spaces instead of usual 2

packages/react/src/modules/slack/components/SlackAuthContainer/styles.css
```diff
@@ -1,28 +1,28 @@
-.rnf-slack-container-container {
-  padding: 20px;
-  border-radius: 8px;
-  border: 1px solid var(--Gray-Gray-100, #e4e8ee);
-  background: var(--Gray-White, #fff);
-  font-family: var(--slack-connect-font-family);
-  font-size: 14px;
-  font-weight: 500;
+.rsk-auth {
```
cjbell, line 1: Nit: same here on spacing

packages/react/src/modules/slack/components/SlackChannelCombobox/SlackChannelCombobox.tsx
```diff
@@ -83,6 +90,16 @@ export const SlackChannelCombobox = ({
     SlackChannelConnection[] | null
   >(null);
 
+  useEffect(() => {
+    if (comboboxListOpen) {
+      // Timeout to allow for the state to update and the component to re-render
+      // when we change the `comboboxListOpen` state upon focus
+      setTimeout(() => {
+        document.getElementById("slack-channel-search")?.focus();
```
cjbell, line 98: I think here it might be best to use:

`comboboxRef.current.querySelector('.rsk-combobox__searchbar__input')?.focus()` so you don't need the id

cjbell (approved):
few nits, nothing blocking and css cleanup looks great 👏

packages/react/src/modules/slack/components/SlackAddChannelInput/styles.css
meryldakin: very weird my tab spacing got changed in vscode somehow

packages/react/src/modules/slack/components/SlackChannelCombobox/SlackChannelCombobox.tsx
meryldakin: is it better to not use the id? the id will exist regardless of this - the reason I'm asking is because i'm getting a typescript error about the `comboboxRef.current` possibly being a never type

connorlindsey: Checking if `ref.current` is defined may fix the typescript error

PR #77, opened 2024-02-26
fix(kno-5338): qa pass 2

meryldakin:
This PR handles Krisna's suggestions from the second QA pass https://www.notion.so/knockapp/SlackKit-QA-doc-a35f9536dc304aa0996c57c492a2cf4e?pvs=4#6e29797f0da4463f865512ce6966cbb0

### DIFF
--- a/packages/react-core/src/modules/slack/hooks/useSlackAuth.ts
+++ b/packages/react-core/src/modules/slack/hooks/useSlackAuth.ts
@@ -9,8 +9,7 @@ const DEFAULT_SLACK_SCOPES = [
   "chat:write",
   "chat:write.public",
   "channels:read",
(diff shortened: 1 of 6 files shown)


packages/react-core/src/modules/slack/hooks/useSlackAuth.ts
```diff
@@ -9,8 +9,7 @@ const DEFAULT_SLACK_SCOPES = [
   "chat:write",
   "chat:write.public",
   "channels:read",
-  "groups:read",
-  "groups:write",
```
meryldakin, line 13: I found out while writing docs this write scope is unnecessary, it allows the bot to create new private channels which we're not using

PR #79, opened 2024-02-26
feat(kno-5344): add i18n translations to slack

meryldakin:
Adds internationalization to the Slack components.


packages/react-core/src/modules/i18n/languages/en.ts
```diff
@@ -11,6 +11,27 @@ const en: I18nContent = {
     unread: "Unread",
     read: "Read",
     unseen: "Unseen",
+    slackConnectChannel: "Connect channel",
```
connorlindsey, line 14: What do you think about scoping translations by module (e.g. feed, slack)? It would organize these objects better. Probably not something we should worry about now though

meryldakin: I think that's a good idea, I kind of wanted to move this through today to get it out sooner because that will involve some changes to the hook. If we think we can move into that in the future I'd rather wait on it.

connorlindsey (approved):

PR #91, opened 2024-03-01
fix: update event format for client status updates

JEverhart383:
This PR adds a second event broadcast on FeedItem status updates in the correct format, e.g. `items.archived`. It also add functionality to the client example to test this new event, which works as expected:
Screenshot 2024-03-01 at 2 51 58 PM
Screenshot 2024-03-01 at 2 52 08 PM
Screenshot 2024-03-01 at 3 01 40 PM

For some reason, the diff make the spacing look funky, but everything looks good if you look at the actual code.

### DIFF
--- a/packages/client/src/clients/feed/feed.ts
+++ b/packages/client/src/clients/feed/feed.ts
@@ -6,312 +6,312 @@ import Knock from "../../knock";
 import { NetworkStatus, isRequestInFlight } from "../../networkStatus";
 
 import {
-	FeedClientOptions,
-	FeedItem,
-	FeedMetadata,
-	FeedResponse,
-	FetchFeedOptions,
+  FeedClientOptions,
+  FeedItem,
+  FeedMetadata,
+  FeedResponse,
+  FetchFeedOptions,
 } from "./interfaces";
 import createStore from "./store";
 import {
-	BindableFeedEvent,
-	FeedEvent,
-	FeedEventCallback,
-	FeedEventPayload,
-	FeedItemOrItems,
-	FeedMessagesReceivedPayload,
-	FeedRealTimeCallback,
-	FeedStoreState,
+  BindableFeedEvent,
+  FeedEvent,
+  FeedEventCallback,
+  FeedEventPayload,
+  FeedItemOrItems,
+  FeedMessagesReceivedPayload,
+  FeedRealTimeCallback,
+  FeedStoreState,
 } from "./types";
 
 export type Status =
-	| "seen"
-	| "read"
-	| "interacted"
-	| "archived"
-	| "unseen"
-	| "unread"
-	| "unarchived";
+  | "seen"
+  | "read"
+  | "interacted"
+  | "archived"
+  | "unseen"
+  | "unread"
+  | "unarchived";
 
 // Default options to apply
 const feedClientDefaults: Pick<FeedClientOptions, "archived"> = {
-	archived: "exclude",
+  archived: "exclude",
 };
 
 const DEFAULT_DISCONNECT_DELAY = 2000;
 
 class Feed {
-	private userFeedId: string;
-	private channel?: Channel;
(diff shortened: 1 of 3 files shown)


packages/client/src/clients/feed/feed.ts
```diff
@@ -6,333 +6,333 @@ import Knock from "../../knock";
 import { NetworkStatus, isRequestInFlight } from "../../networkStatus";
 
 import {
-  FeedClientOptions,
-  FeedItem,
-  FeedMetadata,
-  FeedResponse,
-  FetchFeedOptions,
+	FeedClientOptions,
```
thomaswhyyou, line 9: Jeff is it possible to tweak your editor setting so we can keep things formatted the same / consistently? That will help clean up the diff, but also I think we want to keep all of our public ts/js formatted consistently.

JEverhart383: Yeah, I'm happy to tweak whatever is needed, but I've already go Prettier installed and thought this should pick up the prettier-config file in this repo. If you could also sanity check mere here, but if I look at these two files I can't see any of the formatting differences expressed in the diff. They both look the same to me: https://github.com/knocklabs/javascript/blob/fix/je-update-status-event/packages/client/src/clients/feed/feed.ts
https://github.com/knocklabs/javascript/blob/main/packages/client/src/clients/feed/feed.ts

packages/client/src/clients/feed/feed.ts
```diff
@@ -619,6 +619,10 @@ class Feed {
     // Emit the event that these items had their statuses changed
     // Note: we do this after the update to ensure that the server event actually completed
     this.broadcaster.emit(`items:${type}`, { items });
+
+    // Add this additional event emission to keep event formatting externally consistent
+    // We should make plans to eventually depreated the inconsistent formatting, e.g. items:archived
+    this.broadcaster.emit(`items.${type}`, { items });
     this.broadcastOverChannel(`items:${type}`, { items });
```
thomaswhyyou, line 626:
```suggestion
    // Emit the event that these items had their statuses changed
    // Note: we do this after the update to ensure that the server event actually completed
    this.broadcaster.emit(`items.${type}`, { items });

    // Note: `items.type` format is being deprecated in favor over the `items:type` format, 
    // but emit both formats to make it backward compatible for now.
    this.broadcaster.emit(`items:${type}`, { items });
    this.broadcastOverChannel(`items:${type}`, { items });
```
Just moved the "correct" format up, so the comment re: deprecation reads better to the other ones.

packages/client/src/clients/feed/feed.ts
```diff
@@ -619,6 +619,10 @@ class Feed {
     // Emit the event that these items had their statuses changed
     // Note: we do this after the update to ensure that the server event actually completed
     this.broadcaster.emit(`items:${type}`, { items });
```
thomaswhyyou, line 621: Question: what is the difference between `this.broadcaster.emit()` vs `this.broadcastOverChannel()`? 

Asking because I'm wondering if we need to also `broadcastOverChannel()` with the `items.type` format? Currently it only seems to use the `:` format.

JEverhart383: I believe the `broadcastIOverChannel` has to do with cross-tab communication between feeds, where any event triggers a refetch of the current feed state: https://github.com/knocklabs/javascript/blob/05826b3115df6771f47874cb841b4dfc78f731cd/packages/client/src/clients/feed/feed.ts#L652

I'm not sure if we'd want to update this right now, since it all seems internal

thomaswhyyou: Ah okay, thanks for checking. If it's for internal usage, then yeah we can probably punt that for now and update when we do deprecate the other format.

thomaswhyyou (approved):
🙏

PR #92, opened 2024-03-01
feat(kno-5364): custom onAuthenticationComplete callback for SlackAuthButton

meryldakin:
Adds the custom onAuthenticationComplete callback that accepts the results of the auth attempt for the user to handle.


cjbell (approved):
💯 need to update docs as well please!

kylemcd (approved):

cenobitedk:
Hi! When is this expected to be released? (also asked on Slack, don't know what you see first)

PR #94, opened 2024-03-06
fix(kno-5450): make slack combobox list box opaque

meryldakin:
Makes the list box opaque so that it covers any elements underneath it.


JEverhart383 (approved):
Awesome, TY
Screenshot 2024-03-06 at 11 14 08 AM

PR #97, opened 2024-03-14
feat: support in-app actions in `NotificationCell`

cjbell:
- Adds support for rendering `actions` as buttons in the notification cell
- Adds `onNotificationButtonClick` and `onButtonClick` handlers
- Adds support for sending different template types in the example app

### DIFF
--- a/packages/react/src/modules/feed/components/NotificationCell/NotificationCell.tsx
+++ b/packages/react/src/modules/feed/components/NotificationCell/NotificationCell.tsx
@@ -1,19 +1,31 @@
-import React, { ReactNode, useMemo } from "react";
-import { ContentBlock, FeedItem } from "@knocklabs/client";
-import { Avatar } from "./Avatar";
-import { ArchiveButton } from "./ArchiveButton";
 import {
-  useKnockFeed,
-  useTranslations,
+  ButtonBlock,
+  ButtonSetContentBlock,
+  ContentBlock,
+  FeedItem,
+  MarkdownContentBlock,
+  TextContentBlock,
+} from "@knocklabs/client";
+import {
   formatTimestamp,
   renderNodeOrFallback,
+  useKnockFeed,
+  useTranslations,
 } from "@knocklabs/react-core";
+import React, { ReactNode, useMemo } from "react";
 
+import { Button, ButtonGroup } from "../../../core";
+
+import { ArchiveButton } from "./ArchiveButton";
+import { Avatar } from "./Avatar";
 import "./styles.css";
 
 export interface NotificationCellProps {
   item: FeedItem;
+  // Invoked when the outer container is clicked
   onItemClick?: (item: FeedItem) => void;
+  // Invoked when a button in the notification cell is clicked
+  onButtonClick?: (item: FeedItem, action: ButtonBlock) => void;
   avatar?: ReactNode;
   children?: ReactNode;
   archiveButton?: ReactNode;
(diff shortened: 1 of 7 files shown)


meryldakin (approved):
this looks great!

thomaswhyyou:
Taking a quick look 👀

packages/react/src/modules/feed/components/NotificationCell/NotificationCell.tsx
```diff
@@ -23,91 +35,129 @@ type BlockByName = {
...
+      return maybeNavigateToUrlWithDelay(actionUrl);
+    }, [item, actionUrl, onItemClick, feedClient]);
+
+    const onButtonClickHandler = React.useCallback(
+      (e: React.MouseEvent, button: ButtonBlock) => {
+        feedClient.markAsInteracted(item);
+
+        if (onButtonClick) return onButtonClick(item, button);
+
+        return maybeNavigateToUrlWithDelay(button.action);
```
thomaswhyyou, line 79: Just double checking - is button's `action` expected to be a fully rendered url here? Asking because `ButtonBlock` does not have a `rendered` field, which gave me the impression they aren't liquid rendered.

packages/react/src/modules/feed/components/NotificationCell/NotificationCell.tsx
```diff
@@ -23,91 +35,129 @@ type BlockByName = {
...
+            )}
+
+            {buttonSet && (
+              <div className="rnf-notification-cell__button-group">
+                <ButtonGroup>
+                  {buttonSet.buttons.map((button, i) => (
+                    <Button
+                      variant={i === 0 ? "primary" : "secondary"}
+                      key={button.name}
+                      onClick={(e) => onButtonClickHandler(e, button)}
```
thomaswhyyou, line 138: Not a big deal, but it doesn't seem like `onButtonClickHandler` actually uses the click event.

packages/react/src/modules/feed/components/NotificationCell/NotificationCell.tsx
```diff
@@ -23,91 +35,129 @@ type BlockByName = {
...
+                }}
+              />
+            )}
+
+            {buttonSet && (
+              <div className="rnf-notification-cell__button-group">
+                <ButtonGroup>
+                  {buttonSet.buttons.map((button, i) => (
+                    <Button
+                      variant={i === 0 ? "primary" : "secondary"}
```
thomaswhyyou, line 136:
```suggestion
                      variant={button.name}
```
Alternatively, buttons should have a name field set to either "primary" or "secondary".

packages/react/src/modules/feed/components/NotificationCell/NotificationCell.tsx
cjbell: Yes, it's absolutely the rendered value here. You can see in the feed template renderer for the NotificationEngine where it's doing this. We don't send down `rendered` as it doesn't make sense in the context of these buttons to pass the original liquid template strings and the rendered variables, so we just send down the result.

thomaswhyyou: Got it, thanks for clarifying!

packages/react/src/modules/feed/components/NotificationCell/NotificationCell.tsx
thomaswhyyou: Ah got it, because the variant for `Button` component will still be "secondary" even for the third/tertiary button.

PR #108, opened 2024-04-03
feat: add MessageClient to handle all message update apis

matgar26:


cjbell (approved):

packages/client/src/clients/feed/feed.ts
```diff
@@ -624,13 +622,15 @@ class Feed {
...
+  private async makeBulkStatusUpdate(
+    status: BulkUpdateMessagesInChannelProperties["status"],
+  ) {
     // The base scope for the call should take into account all of the options currently
     // set on the feed, as well as being scoped for the current user. We do this so that
     // we ONLY make changes to the messages that are currently in view on this feed, and not
     // all messages that exist.
     const options = {
-      user_ids: [this.knock.userId],
+      user_ids: [this.knock.userId!],
```
connorlindsey, line 633: Nit: Do we need the `!` here? We should check if the types are correct since before it was defined.

cjbell: We do as the type can be undefined but it will definitely be set here

connorlindsey (approved):

PR #110, opened 2024-04-08
ExpoPushNotificationProvider

matgar26:


packages/react-native/src/modules/push-notifications/ExpoPushNotificationProvider.tsx
```diff
@@ -0,0 +1,236 @@
...
+    return null;
+  }
+
+  try {
+    if (
+      !Constants.expoConfig ||
+      !Constants.expoConfig.extra ||
+      !Constants.expoConfig.extra.eas
+    ) {
+      console.error("Project ID is not defined in the app configuration.");
```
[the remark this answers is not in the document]
connorlindsey: Good idea, could add a `logger` class or something to wrap that. Is the goal here to make it clear to end users when they haven't configured the provider correctly? Would lean towards using `console.warn` over log in that case.

packages/react-native/src/modules/push-notifications/ExpoPushNotificationProvider.tsx
```diff
@@ -0,0 +1,236 @@
...
+      projectId: Constants.expoConfig.extra.eas.projectId,
+    });
+    return token;
+  } catch (error) {
+    console.error("Error getting a push token", error);
+    return null;
+  }
+}
+
+export const ExpoPushNotificationProvider: React.FC<
```
connorlindsey, line 84: Should we include Knock in this provider name like we do with the other providers, including the Slack provider? Since Expo is another organization, the extra namespacing is important imo.

examples/react-native-example/src/App.js
```diff
@@ -10,12 +15,18 @@ export default function App() {
       host={process.env.EXPO_PUBLIC_KNOCK_HOST}
       userId={process.env.EXPO_PUBLIC_KNOCK_USER_ID}
     >
-      <KnockFeedProvider feedId={process.env.EXPO_PUBLIC_KNOCK_FEED_CHANNEL_ID}>
-        <View style={styles.container}>
-          <StatusBar style="auto" />
-          <NotificationFeed />
-        </View>
-      </KnockFeedProvider>
+      <ExpoPushNotificationProvider
```
connorlindsey, line 18: If this provider is using the `useKnockClient`, etc. hooks, it'll need to be a child of `KnockFeedProvider` so it has access to the context.

packages/react-native/src/index.ts
```diff
@@ -1,2 +1,3 @@
 export * from "./modules/feed";
+export * from "./modules/push-notifications";
```
connorlindsey, line 2: Naming nit: Wdyt about `push` for the folder name?

packages/react-native/src/modules/push-notifications/ExpoPushNotificationProvider.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+async function requestPermissionAndGetPushToken(): Promise<Notifications.ExpoPushToken | null> {
+  // Check for device support
+  if (!Device.isDevice) {
+    console.log("Must use physical device for Push Notifications");
+    return null;
+  }
+
+  // Check for permissions
+  const { status: existingStatus } = await Notifications.getPermissionsAsync();
```
connorlindsey, line 51: Might be worth turning this into a function that checks or requests permissions

packages/react-native/package.json
```diff
@@ -40,8 +40,12 @@
     "url": "https://github.com/knocklabs/javascript/issues"
   },
   "peerDependencies": {
-    "react": ">=16.8.0",
-    "react-native": ">=0.60.0"
+    "expo": "~50.0.14",
+    "expo-constants": "^15.4.0",
+    "expo-device": "^5.9.3",
+    "expo-notifications": "^0.27.6",
```
kylemcd, line 46: I think including these as peer dependencies may be incorrect since these are only required when using the expo provider?

packages/react-native/src/modules/push/KnockExpoPushNotificationProvider.tsx
```diff
@@ -0,0 +1,283 @@
...
+async function requestPushPermissionIfNeeded(): Promise<string> {
+  const { status: existingStatus } = await Notifications.getPermissionsAsync();
+  let finalStatus = existingStatus;
+
+  if (existingStatus !== "granted") {
+    const { status } = await Notifications.requestPermissionsAsync();
+    finalStatus = status;
+  }
+
+  return finalStatus;
```
kylemcd, line 71:
```suggestion

  if (existingStatus !== "granted") {
    const { status } = await Notifications.requestPermissionsAsync();
    return status;
  }

  return existingStatus;
```

packages/react-native/src/modules/push/KnockExpoPushNotificationProvider.tsx
```diff
@@ -0,0 +1,283 @@
...
+      if (token?.data) {
+        setExpoPushToken(token.data);
+      }
+    });
+  }
+
+  async function updateKnockMessageStatusFromNotification(
+    notification: Notifications.Notification,
+    status: MessageEngagementStatus,
+  ) {
```
kylemcd, line 147: Should this function return something?

packages/react-native/package.json
matgar26: @kylemcd In this case, how would you recommend declaring these dependencies?

kylemcd (approved):

PR #114, opened 2024-04-10
fix: swap usequery with swr for react 16 support

cjbell:
* Ditches `@tankstack/react-query` for `swr` which supports React 16.11+
* Removes the dependencies and use of `react-query`
* Fixes the react peerDep version in our package.json files

### DIFF
--- a/packages/react-core/src/modules/slack/hooks/useSlackChannels.ts
+++ b/packages/react-core/src/modules/slack/hooks/useSlackChannels.ts
@@ -1,13 +1,15 @@
 import { SlackChannelQueryOptions, useKnockSlackClient } from "..";
-import { SlackChannel } from "@knocklabs/client";
-import { useInfiniteQuery } from "@tanstack/react-query";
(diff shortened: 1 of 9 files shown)


packages/react-core/src/modules/slack/hooks/useSlackChannels.ts
```diff
@@ -19,49 +21,66 @@ type UseSlackChannelOutput = {
   refetch: () => void;
 };
 
+type QueryKey = [key: string, cursor: string] | null;
+
+function getQueryKey(
```
connorlindsey, line 26: Nit: Should we move this generic query key logic to a swr utils file?

connorlindsey (approved):

packages/react-core/src/modules/slack/hooks/useSlackChannels.ts
cjbell: In this case this is very specific to this slack channel payload so no on extracting right now.

PR #115, opened 2024-04-10
fix: use jsx classic runtime to avoid issues on react < 18

cjbell:
When using React < 18 users will run into this error: 

```
Module not found: Error: Can't resolve 'react/jsx-runtime' in '[FILE]'
Did you mean 'jsx-runtime.js'?
BREAKING CHANGE: The request 'react/jsx-runtime' failed to resolve only because it was resolved as fully specified
(probably because the origin is strict EcmaScript Module, e. g. a module with javascript mimetype, a '*.mjs' file, or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
```

The issue is described in this thread: https://github.com/facebook/react/issues/20235

The fix was implemented in React 18, but the only "fixes" that exist if we're using the JSX runtime are to patch webpack configs, which is not great.

Instead, this PR implements an option to use the "classic" JSX runtime, which is the `createElement` API as a workaround so that `react/jsx-runtime` is never referenced. To do so, we also needed to ensure that `import React from "react"` was at the top of all of our JSX components, which is what the `babel-plugin-require-react` does.

I tested this locally in our example apps with a build and everything works

### DIFF
--- a/packages/react-core/vite.config.ts
+++ b/packages/react-core/vite.config.ts
@@ -16,7 +16,12 @@ export default defineConfig(({ mode }) => {
       dts({
         outDir: "dist/types",
       }),
-      react(),
+      react({
+        jsxRuntime: "classic",
+        babel: {
+          plugins: ["react-require"],
+        },
+      }),
       noBundlePlugin({ root: "./src" }),
     ],
     build: {
--- a/.changeset/warm-cooks-work.md
+++ b/.changeset/warm-cooks-work.md
@@ -1,0 +1,6 @@
+---
+"@knocklabs/react-core": patch
+"@knocklabs/react": patch
+---
+
+fix: move away from jsx runtime to support react 16 + 17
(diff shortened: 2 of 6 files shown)


packages/react-core/vite.config.ts
```diff
@@ -16,7 +16,12 @@ export default defineConfig(({ mode }) => {
       dts({
         outDir: "dist/types",
       }),
-      react(),
+      react({
+        jsxRuntime: "classic",
+        babel: {
+          plugins: ["react-require"],
```
connorlindsey, line 22: Is this just for convenience or necessary because of something else in the build? Just curious. I like the idea of having it either way because it'd be easy to forget we have to explicitly import React.

connorlindsey (approved):
Looks good! Will need a changeset as well

packages/react-core/vite.config.ts
cjbell: Convenience because since React 17 everyone got used to never writing `import React` 🙃

PR #124, opened 2024-04-19
fix: feed events should all be items. and fixed types

cjbell:
* feed events should all be `items.${type}` as well as `items:${type}`
* make some generic types interfaces
* allow `FeedEventPayload` to accept a generic

### DIFF
--- a/packages/client/src/clients/feed/feed.ts
+++ b/packages/client/src/clients/feed/feed.ts
@@ -751,7 +751,8 @@ class Feed {
     items: FeedItem[],
   ) {
     // Handle both `items.` and `items:` format for events for compatibility reasons
-    this.broadcaster.emit([`items.${type}`, `items:${type}`], { items });
+    this.broadcaster.emit(`items.${type}`, { items });
+    this.broadcaster.emit(`items:${type}`, { items });
     // Internal events only need `items:`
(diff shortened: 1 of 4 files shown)


packages/client/src/clients/feed/feed.ts
```diff
@@ -750,6 +739,23 @@ class Feed {
...
+      | "all_read"
+      | "all_seen"
+      | "all_archived"
+      | "unread"
+      | "unseen"
+      | "unarchived",
+    items: FeedItem[],
+  ) {
+    // Handle both `items.` and `items:` format for events for compatibility reasons
+    this.broadcaster.emit([`items.${type}`, `items:${type}`], { items });
```
thomaswhyyou, line 754: I assume this does the same thing emitting each one like below?
```
this.broadcaster.emit(`items.${type}`, { items });
this.broadcaster.emit(`items:${type}`, { items });
```

thomaswhyyou (approved):

JEverhart383 (approved):
Awesome, TY

packages/client/src/clients/feed/feed.ts
cjbell: I actually just checked and I was wrong so fixed this!

PR #131, opened 2024-05-08
fix: subscribe to state updates for notifications

kylemcd:
In react strict mode, updates from `useNotifications` was not persisted downstream, so the UI seemed unresponsive as it wasn't getting updates. There is likely an underlying issue that we need to investigate further, but using zustand's [transient updates](https://github.com/pmndrs/zustand?tab=readme-ov-file#transient-updates-for-often-occurring-state-changes) pattern seems to work for both react strict mode and a production build.

### DIFF
--- a/packages/react-core/src/modules/feed/hooks/useNotifications.ts
+++ b/packages/react-core/src/modules/feed/hooks/useNotifications.ts
@@ -1,5 +1,5 @@
 import Knock, { Feed, FeedClientOptions } from "@knocklabs/client";
-import { useMemo, useRef } from "react";
+import { useMemo, useRef  } from "react";
 
 function useNotifications(
   knock: Knock,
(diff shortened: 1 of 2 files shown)


packages/react-core/src/modules/feed/hooks/useNotifications.ts
```diff
@@ -13,12 +13,13 @@ function useNotifications(
       feedClientRef.current.dispose();
     }
 
-    const feedClient = knock.feeds.initialize(feedChannelId, options);
+    feedClientRef.current = knock.feeds.initialize(feedChannelId, options);
+    feedClientRef.current.listenForUpdates();
+    feedClientRef.current.store.subscribe((t) =>
+      feedClientRef?.current?.store.setState(t),
+    );
```
cjbell, line 20: mentioned in slack but wondering if it's worth wrapping this in if (!process.env.NODE_ENV || process.env.NODE_ENV === 'development')`?

cjbell (approved):

PR #134, opened 2024-05-14
fix: ensure that options are stable between re-renders for useAuthenticatedKnockClient

cjbell:
- Uses a shallow comparison and a memo for `options` passed into the `useAuthenticatedKnockClient` hook to ensure the options are stable between re-renders
- Adds a test for the hook (!)

### DIFF
--- a/packages/react-core/src/modules/core/hooks/useAuthenticatedKnockClient.ts
+++ b/packages/react-core/src/modules/core/hooks/useAuthenticatedKnockClient.ts
@@ -1,5 +1,6 @@
 import Knock, { AuthenticateOptions, KnockOptions } from "@knocklabs/client";
-import React, { useMemo } from "react";
+import React from "react";
+import shallow from "zustand/shallow";
 
 function authenticateWithOptions(
   knock: Knock,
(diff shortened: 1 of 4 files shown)


packages/react-core/src/modules/core/hooks/useAuthenticatedKnockClient.ts
```diff
@@ -13,15 +14,26 @@ function authenticateWithOptions(
...
 ) {
-  const knockRef = React.useRef<Knock | null>();
+  const knockRef = React.useRef<Knock | undefined>();
+  const optionsRef = React.useRef<AuthenticatedKnockClientOptions>();
+
+  // Shallow compare options so that we ensure that we have a stable
+  // set of options between re-renders.
+  const stableOptions = React.useMemo(() => {
+    const currentOptions = optionsRef.current || {};
+    return shallow(options, currentOptions) ? currentOptions : options;
```
kylemcd, line 33: Would `useShallow` make more sense here? https://docs.pmnd.rs/zustand/guides/prevent-rerenders-with-use-shallow

packages/react-core/test/core/useAuthenticatedKnockClient.test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    const authenticateSpy = vi.spyOn(result.current, "authenticate");
+
+    expect(result.current).toBeInstanceOf(Knock);
+    expect(result.current.userId).toEqual("user_123");
+
+    rerender(defaultProps);
+
+    expect(authenticateSpy).not.toBeCalled();
+  });
```
kylemcd, line 51: Should we have a test case that a new instance is created when userId changes?

packages/react-core/src/modules/core/hooks/useAuthenticatedKnockClient.ts
```diff
@@ -13,15 +14,26 @@ function authenticateWithOptions(
...
+  }, [options]);
 
-  return useMemo(() => {
+  return React.useMemo(() => {
```
connorlindsey, line 36: Is this just a style change or does this actually have an impact?

connorlindsey (approved):

PR #138, opened 2024-05-16
fix: clean up eslint configs and warnings and ensure lint runs in ci

connorlindsey:
- Cleans up our eslint dependencies and configs
- Fixes or disables existing eslint warnings
- Runs lint in ci


packages/client/src/clients/objects/index.ts
```diff
@@ -12,6 +12,7 @@ type SetChannelDataInput = {
   objectId: string;
   collection: string;
   channelId: string;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
connorlindsey, line 15: This is the main rule I'm not sure how we want to handle. We use `any` in a number of places, especially as defaults for generics. I think we can use `unknown` as well which will require users to pass in a type to the generic. We could also disable this rule and allow explicit any while still disallowing implicit any.

packages/eslint-config/library.js
```diff
@@ -4,8 +4,14 @@ const project = resolve(process.cwd(), "tsconfig.json");
 
 /** @type {import("eslint").Linter.Config} */
 module.exports = {
-  extends: ["eslint:recommended", "prettier", "eslint-config-turbo"],
+  extends: [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "prettier",
+    "eslint-config-turbo",
```
thomaswhyyou, line 11: Is it fair to say `eslint-config-turbo` is relevant for every package in javascript repo?

Asking because we want to share this `eslint-config` package in control dashboard, and turborepo lint rules aren't applicable there. But when I run eslint with this config, only warnings we are getting is "turbo/no-undeclared-env-vars", so maybe it's easier to just keep this in and then turn it off there.

connorlindsey: Yeah, I think we should keep `eslint-config-turbo` here especially because it can point out ways we're not using turborepo correctly. If it's easy to override the config from the dashboard, let's do that. Otherwise I can take a look at not including it when we publish or have a separate config for `monorepo-library` (or something along those lines)

.github/workflows/cicd.yml
```diff
@@ -23,5 +23,14 @@ jobs:
       - name: Install dependencies
         run: yarn --immutable
 
+      - name: Build packages
+        run: yarn build:packages
```
thomaswhyyou, line 27: Is `build:packages` required for the new steps we added below?

packages/eslint-config/next.js
```diff
@@ -6,10 +6,12 @@ const project = resolve(process.cwd(), "tsconfig.json");
 module.exports = {
   extends: [
     "eslint:recommended",
```
thomaswhyyou, line 8: What's the difference between this file and library.js? And should they be the same?

packages/types/tsup.config.ts
```diff
@@ -1,15 +0,0 @@
-import { defineConfig } from 'tsup';
```
thomaswhyyou, line 1: What's this file?

packages/client/src/clients/objects/index.ts
thomaswhyyou: A few thoughts:
- Any places where are using `any`, but what we really mean is any key/value object, it looks like we have `GenericData` we can use?
- Any places where are using `any` but there's more we can type, we probably should eventually. So maybe slap a TODO in those places?
  - Separately, I'm curious how auto SDK generation efforts will come into play here. Because I imagine when we auto generate node sdk, there will be some types generated as well, and I imagine we would want/be able to re-use those? So maybe this is all an temporary issue.

thomaswhyyou (commented):
Just a few questions out of curiosity, but overall looks reasonable to me.

packages/eslint-config/next.js
connorlindsey: The main difference is that this has the Vercel Next.js eslint rules and is meant to be used in Next apps

packages/types/tsup.config.ts
connorlindsey: It was for a bundler I was testing with the types package but didn't end up needing. Must've missed deleting it when I added the types package, so removing now as its not actually used.

.github/workflows/cicd.yml
connorlindsey: Tests need it so that each package has the latest version of its dependencies (e.g. React has the latest build of React-core). If I understand correctly, turbo should automatically take care of that because in `turbo.json` we list build as a dependency of test, but I added here just to make it clear we're doing a build and that it gets listed as a separate step.

packages/client/src/clients/objects/index.ts
connorlindsey: That's a really good point. I'll replace all of these with `GenericData` as that's what we want, then we just have a single eslint disable comment.

connorlindsey: Update: Great suggestion from Thomas to use `GenericData` means we really only have 1 place where have an explicit any

thomaswhyyou (approved):
🫡

PR #143, opened 2024-05-17
fix: ensure feed renders work correctly in next 14.2

cjbell:
* Adds a new next 14.2 app router app
* Fixes issue with re-renders not binding on real-time events


cjbell:
Closing as I don't want to figure out all of the build nonsense right now.

PR #147, opened 2024-05-22
Update dependencies and a fix to our example app

matgar26:
Updating how we declare our Expo peer dependencies to be more flexible with what versions our clients can use. Also added a fix into our Example App dealing with our NotificationFeed.


examples/react-native-example/src/components/NotificationFeed.js
```diff
@@ -10,17 +10,31 @@ export default function NotificationFeed() {
...
+    feedClient.on("messages.new", handleNewMessage);
+    feedClient.on("items.received.*", handleItemsReceived);
+
+    return () => {
+      // Unsubscribe from events
+      feedClient.off("messages.new", handleNewMessage);
+      feedClient.off("items.received.*", handleItemsReceived);
 
-    return () => teardown();
+      // Call teardown if it is a function
```
connorlindsey, line 33: Nit: These comments feel unnecessary since it's clear from the code what is happening

connorlindsey (approved):
Looks great!

PR #152, opened 2024-05-29
fix: remove event stop propogation from outside click hook

kylemcd:
### Description
A customer was experiencing an issue where when the notification feed popover was open, clicking a link outside of the popover would cause the page to hard refresh when using `react-router-dom`. Removing `event.stopPropogation()` from the `useComponentVisible` hook fixed this issue. 

### Tasks
[KNO-6065](https://linear.app/knock/issue/KNO-6065/[js-sdk]-investigate-event-propagation-issue-with-react-popover)

### DIFF
--- a/.changeset/rich-foxes-tease.md
+++ b/.changeset/rich-foxes-tease.md
@@ -1,0 +1,5 @@
+---
+"@knocklabs/react": patch
+---
+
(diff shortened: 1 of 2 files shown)


cjbell (approved):
💯

PR #168, opened 2024-08-01
feat: react native in app components

matgar26:


connorlindsey:
We use a tool to keep dependencies in sync between packages. You can run `yarn manypkg fix` to fix any discrepancies and that'll fix CI.

packages/react-native/src/theme/theme.ts
```diff
@@ -0,0 +1,93 @@
...
+    knock0: 11,
+    knock1: 12,
+    knock2: 14,
+    knock3: 16,
+    knock4: 18,
+    knock5: 20,
+    knock6: 24,
+    knock7: 30,
+    knock8: 36,
+    knock9: 48,
```
cjbell, line 27: Nit: can we remove the `knock` prefix here please. This is already under our specific package so is superfluous

packages/react-native/src/modules/feed/components/NotificationFeedComponents/NotificationFeedCell/NotificationFeedCell.tsx
```diff
@@ -0,0 +1,246 @@
...
+                ]}
+              />
+              {resolvedStyle.showAvatarView && actor && isUser(actor) && (
+                <View>
+                  <AvatarView name={actor.name} src={actor.avatar} />
+                </View>
+              )}
+            </View>
+            <View style={layoutStyles.contentContainer}>
+              {item.blocks.map((block, index) => {
```
cjbell, line 171: I think one thing we're missing here right now is support for the `action_url` block, which is a text block that contains the action that should be invoked on tap of the notification cell.

cjbell (commented):
this looks awesome matt! tiny comments

packages/client/src/clients/feed/feed.ts
```diff
@@ -410,6 +410,53 @@ class Feed {
     return result;
   }
 
+  async markAllReadAsArchived() {
+    // Note: there is the potential for a race condition here because the bulk
+    // update is an async method, so if a new message comes in during this window before
+    // the update has been processed we'll effectively reset the `unseen_count` to be what it was.
+    const { items, ...state } = this.store.getState();
+    console.log("[Feed] Items", items);
```
cjbell, line 418: 👀 can you remove all of these console.logs please

PR #184, opened 2024-08-15
feat: record interaction metadata for button and container clicks

cjbell:
* Record interaction metadata on button click and cell click
* Add support for recording metadata in the client and feed methods

KNO-6669

### DIFF
--- a/packages/client/src/clients/feed/feed.ts
+++ b/packages/client/src/clients/feed/feed.ts
@@ -286,7 +286,10 @@ class Feed {
     return this.makeStatusUpdate(itemOrItems, "unread");
   }
 
-  async markAsInteracted(itemOrItems: FeedItemOrItems) {
+  async markAsInteracted(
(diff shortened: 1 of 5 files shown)


packages/client/src/clients/feed/feed.ts
```diff
@@ -583,12 +586,17 @@ class Feed {
   private async makeStatusUpdate(
     itemOrItems: FeedItemOrItems,
     type: MessageEngagementStatus | "unread" | "unseen" | "unarchived",
+    metadata?: Record<string, string>,
```
kylemcd, line 589: Maybe a nit but should we consider changing this this third function param to an object that can contain a meta data var? Starts to get smelly after 2 function parameters. Same for the markAsInteracted.

kylemcd (approved):

connorlindsey (approved):

PR #219, opened 2024-09-23
feat: add basic in app channel and message clients to fetch messages

connorlindsey:
- Adds basic in app channel client and message clients to support fetching messages from the API
- Adds demo application that polls for messages and displays the stored messages

### DIFF
--- a/packages/client/src/clients/in-app/message-client.ts
+++ b/packages/client/src/clients/in-app/message-client.ts
@@ -1,0 +1,98 @@
+import Knock from "../../knock";
+import { NetworkStatus, isRequestInFlight } from "../../networkStatus";
+
+import { InAppChannelClient } from "./channel-client";
+import {
+  FetchInAppMessagesOptions,
+  InAppMessageClientOptions,
+  InAppMessagesQueryInfo,
+} from "./types";
+
+const defaultOptions: Pick<InAppMessageClientOptions, "archived"> = {
+  archived: "exclude",
+};
+
+/**
+ * Manages realtime connection to in app messages service.
+ */
+export class InAppMessageClient {
+  private knock: Knock;
+  private defaultOptions: InAppMessageClientOptions;
+
+  constructor(
+    readonly channelClient: InAppChannelClient,
+    readonly schema: string,
+    options: InAppMessageClientOptions,
+  ) {
+    this.knock = channelClient.knock;
+    this.defaultOptions = { ...defaultOptions, ...options };
+  }
+
+  async fetch(options: FetchInAppMessagesOptions = {}) {
+    // TODO: Create better abstraction for reading from / writing to store
+    const { networkStatus } = this.channelClient.store.state;
+
+    // If there's an existing request in flight, then do nothing
+    if (isRequestInFlight(networkStatus)) {
+      return;
+    }
+
+    // Set the loading type based on the request type it is
+    this.channelClient.store.setState((state) => ({
+      ...state,
+      networkStatus: options.__loadingType ?? NetworkStatus.loading,
+    }));
+
+    const queryParams = {
+      ...this.defaultOptions,
+      ...options,
+      // Unset options that should not be sent to the API
+      __loadingType: undefined,
+      __fetchSource: undefined,
+    };
+
+    const queryParamsString = JSON.stringify(queryParams);
+
+    const result = await this.knock.client().makeRequest({
+      method: "GET",
+      url: `/v1/users/${this.knock.userId}/in-app-messages/${this.channelClient.channelId}`,
+      params: queryParams,
+    });
+
+    if (result.statusCode === "error" || !result.body) {
+      this.channelClient.store.setState((state) => ({
+        ...state,
+        networkStatus: NetworkStatus.error,
+      }));
+
+      return {
+        status: result.statusCode,
+        data: result.error || result.body,
+      };
+    }
+
+    const response: InAppMessagesQueryInfo = {
+      items: result.body.items,
+      pageInfo: result.body.page_info,
+    };
+
+    // TODO: Update queries state page info based on before/after like in feed.ts
+    this.channelClient.store.setState((state) => {
+      return {
+        ...state,
+        // Store new messages in shared store
+        messages: response.items.reduce((messages, message) => {
+          messages[message.id] = message;
+          return messages;
+        }, state.messages),
+        // Store query results
+        queries: {
+          ...state.queries,
+          [queryParamsString]: response,
+        },
+      };
+    });
+
+    return { data: response, status: result.statusCode };
+  }
+}
--- a/packages/client/src/clients/in-app/store.ts
+++ b/packages/client/src/clients/in-app/store.ts
@@ -1,0 +1,19 @@
+import { Store } from "@tanstack/store";
+
+import { NetworkStatus } from "../../networkStatus";
+
+import { InAppStoreState } from "./types";
+
+export type InAppStore = Store<
+  InAppStoreState,
+  (cb: InAppStoreState) => InAppStoreState
+>;
+
+export function createStore() {
+  return new Store<InAppStoreState>({
+    networkStatus: NetworkStatus.ready,
+    loading: false,
+    messages: {},
+    queries: {},
+  });
+}
--- a/packages/client/src/clients/in-app/types.ts
+++ b/packages/client/src/clients/in-app/types.ts
@@ -1,0 +1,81 @@
+import { GenericData, PageInfo } from "@knocklabs/types";
+
+import { NetworkStatus } from "../../networkStatus";
+import { NotificationSource } from "../messages/interfaces";
+
+export interface CommonContentField {
+  key: string;
+  label: string;
+  rendered: string;
+  type: "text" | "markdown" | "boolean" | "textarea";
+  value: string;
+}
+
+export interface ButtonContentField {
+  key: string;
+  type: "button";
+  text: Exclude<CommonContentField, "key | type"> & {
+    key: "text";
+    type: "text";
+  };
+  action: Exclude<CommonContentField, "key | type"> & {
+    key: "action";
+    type: "text";
+  };
+}
+
+export type ContentField = CommonContentField | ButtonContentField;
+
+export interface InAppMessage<T = GenericData> {
+  __cursor: string;
+  id: string;
+  message_type: string;
+  schema_variant_key: string;
+  schema_version: string;
+  priority: number;
+  content: Record<string, ContentField>;
+  data: T | null;
+  inserted_at: string;
+  updated_at: string;
+  seen_at: string | null;
+  read_at: string | null;
+  interacted_at: string | null;
+  archived_at: string | null;
+  link_clicked_at: string | null;
+  source: NotificationSource;
+}
+
+export interface InAppMessagesQueryInfo {
+  items: InAppMessage[];
+  pageInfo: PageInfo;
+}
+
+export interface InAppStoreState {
+  networkStatus: NetworkStatus;
+  loading: boolean;
+  messages: Record<string, InAppMessage>;
+  queries: Record<string, InAppMessagesQueryInfo>;
+}
+
+export interface InAppMessageClientOptions {
+  order?: "asc" | "desc";
+  before?: string;
+  after?: string;
+  page_size?: number;
+  status?: "unread" | "read" | "unseen" | "seen" | "all";
+  // Optionally scope all requests to a particular tenant
+  tenant?: string;
+  // Optionally scope to notifications with any tenancy or no tenancy
+  has_tenant?: boolean;
+  // Optionally scope to notifications with any of the categories provided
+  workflow_categories?: string[];
+  // Optionally scope to a given archived status (defaults to `exclude`)
+  archived?: "include" | "exclude" | "only";
+  // Optionally scope all notifications that contain this argument as part of their trigger payload
+  trigger_data?: GenericData;
+}
+
+export type FetchInAppMessagesOptions = {
+  __loadingType?: NetworkStatus.loading | NetworkStatus.fetchMore;
+  __fetchSource?: "socket" | "http";
+} & InAppMessageClientOptions;
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -71,6 +71,7 @@
   "dependencies": {
     "@babel/runtime": "^7.25.6",
     "@knocklabs/types": "workspace:^",
+    "@tanstack/store": "^0.5.5",
     "@types/phoenix": "^1.5.4",
     "axios": "^1.7.4",
     "axios-retry": "^3.1.9",
--- a/packages/client/src/clients/in-app/channel-client.ts
+++ b/packages/client/src/clients/in-app/channel-client.ts
@@ -1,0 +1,18 @@
+import Knock from "../../knock";
+
+import { InAppStore, createStore } from "./store";
+
+/**
+ * Manages the configuration for an in app channel.
+ * Stores all fetched messages to support optimistic updates.
+ */
+export class InAppChannelClient {
+  public store: InAppStore;
+
+  constructor(
+    readonly knock: Knock,
+    readonly channelId: string,
+  ) {
+    this.store = createStore();
+  }
+}
--- a/packages/client/src/clients/in-app/index.ts
+++ b/packages/client/src/clients/in-app/index.ts
@@ -1,0 +1,3 @@
+export * from "./channel-client";
+export * from "./message-client";
+export * from "./types";
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -12,6 +12,7 @@ export * from "./clients/slack/interfaces";
 export * from "./clients/users";
 export * from "./clients/users/interfaces";
 export * from "./clients/messages";
+export * from "./clients/in-app";
 export * from "./clients/messages/interfaces";
 export * from "./networkStatus";
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -4760,6 +4760,7 @@ __metadata:
     "@babel/preset-typescript": "npm:^7.16.7"
     "@babel/runtime": "npm:^7.25.6"
     "@knocklabs/types": "workspace:^"
+    "@tanstack/store": "npm:^0.5.5"
     "@types/jsonwebtoken": "npm:^9.0.5"
     "@types/phoenix": "npm:^1.5.4"
     "@typescript-eslint/eslint-plugin": "npm:^6.20.0"
@@ -7105,6 +7106,13 @@ __metadata:
   dependencies:
     defer-to-connect: "npm:^2.0.1"
   checksum: 10c0/4629d2fbb2ea67c2e9dc03af235c0991c79ebdddcbc19aed5d5732fb29ce01c13331e9b1a491584b9069bd6ecde6581dcbf871f11b7eefdebbab34de6cf2197e
+  languageName: node
+  linkType: hard
+
+"@tanstack/store@npm:^0.5.5":
+  version: 0.5.5
+  resolution: "@tanstack/store@npm:0.5.5"
+  checksum: 10c0/d06ba882f4196c24ffd1f72f31f3b8939e9fabe642bb03800451cf62b621a60c863c3b4a876384fd13fe3c70bfcae5d230243dba90faa45c15fdaad89e2787f1
   languageName: node
   linkType: hard
 


examples/in-app-messaging-example/src/App.tsx
```diff
@@ -0,0 +1,54 @@
...
+} from "@knocklabs/client";
+import { useEffect, useState } from "react";
+
+const knockClient = new Knock(import.meta.env.VITE_KNOCK_API_KEY!, {
+  host: import.meta.env.VITE_KNOCK_HOST,
+});
+
+knockClient.authenticate(import.meta.env.VITE_KNOCK_USER_ID!);
+
+const inAppChannelClient = new InAppChannelClient(
```
connorlindsey, line 13: Will replace with `<KnockInAppChannelProvider />`

examples/in-app-messaging-example/src/App.tsx
```diff
@@ -0,0 +1,54 @@
...
+  knockClient,
+  import.meta.env.VITE_KNOCK_CHANNEL_ID,
+);
+
+const inAppMessageClient = new InAppMessageClient(
```
connorlindsey, line 18: Will replace with `useInAppMessages()`

examples/in-app-messaging-example/README.md
```diff
@@ -0,0 +1,3 @@
+# Knock In App Messaging Example
```
thomaswhyyou, line 1:
```suggestion
# Knock In-App Message Example
```
Nit, but I wonder if we just want to just align our wording to "in-app message" (w/o -ing) everywhere including file/dir names.

examples/in-app-messaging-example/src/index.css
```diff
@@ -0,0 +1,68 @@
...
+  min-width: 320px;
+  padding: 1rem 2rem;
+}
+
+h1 {
+  font-size: 3.2em;
+  line-height: 1.1;
+}
+
+button {
```
thomaswhyyou, line 38: Question: Is it correct to assume all/most of the styles in this css will be moved to actual component css eventually? i.e. `banner__title`, `banner__button`, etc.

packages/client/src/clients/in-app/types.ts
```diff
@@ -0,0 +1,81 @@
...
+export interface InAppMessageClientOptions {
+  order?: "asc" | "desc";
+  before?: string;
+  after?: string;
+  page_size?: number;
+  status?: "unread" | "read" | "unseen" | "seen" | "all";
+  // Optionally scope all requests to a particular tenant
+  tenant?: string;
+  // Optionally scope to notifications with any tenancy or no tenancy
+  has_tenant?: boolean;
```
thomaswhyyou, line 69: Not yet supported, and also these filter params might/will look different. I have a PR I'm finishing up here: https://github.com/knocklabs/switchboard/pull/2912

packages/client/src/clients/in-app/types.ts
```diff
@@ -0,0 +1,81 @@
...
+  messages: Record<string, InAppMessage>;
+  queries: Record<string, InAppMessagesQueryInfo>;
+}
+
+export interface InAppMessageClientOptions {
+  order?: "asc" | "desc";
+  before?: string;
+  after?: string;
+  page_size?: number;
+  status?: "unread" | "read" | "unseen" | "seen" | "all";
```
thomaswhyyou, line 65: `engagement_status` as a list of statuses, and also double check on accept values.

packages/client/src/clients/in-app/types.ts
```diff
@@ -0,0 +1,81 @@
...
+  // Optionally scope all requests to a particular tenant
+  tenant?: string;
+  // Optionally scope to notifications with any tenancy or no tenancy
+  has_tenant?: boolean;
+  // Optionally scope to notifications with any of the categories provided
+  workflow_categories?: string[];
```
thomaswhyyou, line 71: Looks like we are missing a workflow_key param. Worth referencing this PR: https://github.com/knocklabs/switchboard/pull/2840/files#diff-5d7a326024a0861c8c3ecbc166b4e8e19fe52ad881881ee57910c20f7c92e4f9R1

packages/client/src/clients/in-app/message-client.ts
```diff
@@ -0,0 +1,98 @@
...
+import { InAppChannelClient } from "./channel-client";
+import {
+  FetchInAppMessagesOptions,
+  InAppMessageClientOptions,
+  InAppMessagesQueryInfo,
+} from "./types";
+
+const defaultOptions: Pick<InAppMessageClientOptions, "archived"> = {
+  archived: "exclude",
+};
```
thomaswhyyou, line 13: Fwiw, by default we already set to "exclude" in the backend so not sure if this is strictly necessary.

packages/client/src/clients/in-app/message-client.ts
```diff
@@ -0,0 +1,98 @@
...
+ * Manages realtime connection to in app messages service.
+ */
+export class InAppMessageClient {
+  private knock: Knock;
+  private defaultOptions: InAppMessageClientOptions;
+
+  constructor(
+    readonly channelClient: InAppChannelClient,
+    readonly messageType: string,
+    options: InAppMessageClientOptions,
```
thomaswhyyou, line 25:
```suggestion
    options: InAppMessageClientOptions = {},
```
Nit, can fall back to an empty option obj, so you don't have to pass in `{}` at a constructor call?

packages/client/src/clients/in-app/store.ts
```diff
@@ -0,0 +1,19 @@
...
+  InAppStoreState,
+  (cb: InAppStoreState) => InAppStoreState
+>;
+
+export function createStore() {
+  return new Store<InAppStoreState>({
+    networkStatus: NetworkStatus.ready,
+    loading: false,
+    messages: {},
+    queries: {},
```
thomaswhyyou, line 17: Can i double check / clarify that this is the global store that is meant to store all messages state? If so, i'm a bit confused why it includes `networkStatus`, `loading` and `queries` - aren't they all part of the InAppMessageClient concerns?

thomaswhyyou (commented):
Took a first pass, and left some comments/questions!

packages/client/src/clients/in-app/types.ts
connorlindsey: Thanks for the heads up!

packages/client/src/clients/in-app/message-client.ts
connorlindsey: Sounds good! Was following the existing `feed` but we can simplify this

examples/in-app-messaging-example/README.md
connorlindsey: Yeah, that sounds good to me!

packages/client/src/clients/in-app/store.ts
connorlindsey: Yeah, this is the global store. I should move `networkStatus` and `loading` inside of the individual queries. 

I'd lean towards keeping the queries in the top level store in case there are duplicate queries and since it is a bit simpler to have a single store.

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,92 @@
+import { GenericData, PageInfo } from "@knocklabs/types";
+
+import { NetworkStatus } from "../../networkStatus";
+import { NotificationSource } from "../messages/interfaces";
+
+export interface CommonContentField {
+  key: string;
+  label: string;
```
thomaswhyyou, line 8: I ended up removing the `label` field, so i think safe to remove it here.

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  rendered: string;
```
thomaswhyyou, line 9: Should be boolean for a boolean field type.

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  | "uninteracted";
+
+export interface InAppMessageClientOptions {
+  order?: "asc" | "desc";
+  before?: string;
+  after?: string;
+  page_size?: number;
+  engagement_status?: EngagementStatus[];
+  // Optionally scope all requests to a particular tenant
+  tenant_id?: string;
```
thomaswhyyou, line 76: This is also a list of tenant ids, similar to engagement_status.

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  // Optionally scope to notifications with any tenancy or no tenancy
+  with_null_tenant?: boolean;
+  // Optionally scope to notifications from the given workflow
+  workflow_key?: string;
```
thomaswhyyou, line 80: Also a list of workflow keys.

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  items: InAppMessage[];
+  pageInfo?: PageInfo;
+}
+
+export interface InAppStoreState {
+  messages: Record<string, InAppMessage>;
+  queries: Record<string, InAppMessagesQueryInfo>;
+}
+
+export type EngagementStatus =
```
thomaswhyyou, line 59:
```suggestion
export type InAppMessageEngagementStatus =
```

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,92 @@
...
+    key: "text";
+    type: "text";
+  };
+  action: Exclude<CommonContentField, "key | type"> & {
+    key: "action";
+    type: "text";
+  };
+}
+
+export type ContentField = CommonContentField | ButtonContentField;
```
thomaswhyyou, line 27: I wonder if these should be prefixed with "InAppMessage" also. Not a big deal if they aren't exposed outside this package though.

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  workflow_categories?: string[];
+  // Optionally scope to a given archived status (defaults to `exclude`)
+  archived?: "include" | "exclude" | "only";
+  // Optionally scope all notifications that contain this argument as part of their trigger payload
+  trigger_data?: GenericData;
+}
+
+export type FetchInAppMessagesOptions = {
+  __loadingType?: NetworkStatus.loading | NetworkStatus.fetchMore;
+  __fetchSource?: "socket" | "http";
```
thomaswhyyou, line 91: Are these Axios specific / related things?

packages/client/src/clients/in-app-message/channel-client.ts
```diff
@@ -0,0 +1,18 @@
+import Knock from "../../knock";
+
+import { InAppStore, createStore } from "./store";
```
thomaswhyyou, line 3: Nit, feel free to ignore - but I feel like these can just be brought into internal types/funcs or inlined into this module.

packages/client/src/clients/in-app-message/message-client.ts
```diff
@@ -0,0 +1,109 @@
...
+    const queryState = this.channelClient.store.state.queries[queryKey] ?? {
+      items: [],
+      loading: false,
+      networkStatus: NetworkStatus.ready,
+    };
+    const networkStatus = queryState.networkStatus;
+
+    // If there's an existing request in flight, then do nothing
+    if (networkStatus && isRequestInFlight(networkStatus)) {
+      return;
```
thomaswhyyou, line 45: Is okay to return `undefined` here though?

packages/client/src/clients/in-app-message/message-client.ts
```diff
@@ -0,0 +1,109 @@
...
+    // TODO: Create better abstraction for reading from / writing to store
+    const queryParams = {
+      ...this.defaultOptions,
+      ...options,
+      // Unset options that should not be sent to the API
+      __loadingType: undefined,
+      __fetchSource: undefined,
+    };
+
+    const queryKey = `${this.messageType}-${JSON.stringify(queryParams)}`;
```
thomaswhyyou, line 35: I'm wondering if we can just use an encoded url path + query params for a query key, vs this one-off format? For example, "/v1/users/thomas/in-app-messages/channel-id/banner?foo=bar". 

Maybe we are okay with the proposed format, but at least the query url path will include a user id, which means the cache is scoped to an individual recipient. Could also be a TODO thing.

packages/client/src/clients/in-app-message/message-client.ts
```diff
@@ -0,0 +1,109 @@
...
+        }, state.messages),
+        // Store query results
+        queries: {
+          ...state.queries,
+          [queryKey]: queryInfo,
+        },
+      };
+    });
+
+    return { data: queryInfo, status: result.statusCode };
```
thomaswhyyou, line 107: `loading` and `networkStatus` included as part of `data` feels kinda weird? I would expect `data` to be just the response payload.

thomaswhyyou (commented):
Took another pass! Most of the comments in types.ts are small / naming / typing related, and outside of that just a couple comments/questions in message-client.ts re: how we handle and structure data.

Happy to chat through if easier.

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,92 @@
...
+export interface InAppMessageClientOptions {
+  order?: "asc" | "desc";
+  before?: string;
+  after?: string;
+  page_size?: number;
+  engagement_status?: EngagementStatus[];
+  // Optionally scope all requests to a particular tenant
+  tenant_id?: string;
+  // Optionally scope to notifications with any tenancy or no tenancy
+  with_null_tenant?: boolean;
```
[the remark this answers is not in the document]
connorlindsey: Sounds good! Will also update tenant_id to take a string or string[]

packages/client/src/clients/in-app-message/channel-client.ts
connorlindsey: I'm planning on adding a few more methods to make writing selectors easier, but if I don't end up adding much more, then I agree it'd be good to consolidate the files.

packages/client/src/clients/in-app-message/message-client.ts
connorlindsey: Agreed! I actually changed that in the next pr in the stack

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,92 @@
...
+
+import { NetworkStatus } from "../../networkStatus";
+import { NotificationSource } from "../messages/interfaces";
+
+export interface CommonContentField {
+  key: string;
+  label: string;
+  rendered: string;
+  type: "text" | "markdown" | "boolean" | "textarea";
+  value: string;
```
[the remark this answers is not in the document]
connorlindsey: Good catch, I'll break out boolean. I think it's fine to keep text/markdown/textarea a shared type

packages/client/src/clients/in-app-message/types.ts
connorlindsey: Same question here, is it always a list or `string | string[]`?

packages/client/src/clients/in-app-message/types.ts
connorlindsey: They are exported, so I'll add the prefix!

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,92 @@
...
+  archived_at: string | null;
+  link_clicked_at: string | null;
+  source: NotificationSource;
+}
+
+export interface InAppMessagesQueryInfo {
+  networkStatus: NetworkStatus;
+  loading: boolean;
+  items: InAppMessage[];
+  pageInfo?: PageInfo;
```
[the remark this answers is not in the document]
connorlindsey: Good idea. I'll make it a data field and fix the types

packages/client/src/clients/in-app-message/types.ts
connorlindsey: No, they're copied from options used in `feed.ts` to handle some conditional logic in fetch like whether to emit `"items.received.realtime"` or `"items.received.page"`. Fairly confident we'll need them, so would lean towards keeping them in.

packages/client/src/clients/in-app-message/types.ts
thomaswhyyou: Originally it was a list, but I just made a tweak in [this PR](https://github.com/knocklabs/switchboard/pull/2912/files#r1772320567) so that we can accept a single value or a list of strings.

connorlindsey:
* **#221** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/221?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#220** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/220?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#219** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/219?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

packages/client/src/clients/in-app-message/message-client.ts
connorlindsey: Yeah, I think it's fine to return nothing here. We do the same thing in the feeds client.

packages/client/src/clients/in-app-message/message-client.ts
connorlindsey: Yeah, I like that for the query key so that it's more unique and probably easier to work with if we need to extract the values from it! I'll make a note to update it

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,103 @@
...
+  | "link_unclicked"
+  | "interacted"
+  | "uninteracted";
+
+export interface InAppMessageClientOptions {
+  order?: "asc" | "desc";
+  before?: string;
+  after?: string;
+  page_size?: number;
+  engagement_status?: InAppMessageEngagementStatus[];
```
thomaswhyyou, line 87:
```suggestion
  engagement_status?: InAppMessageEngagementStatus | InAppMessageEngagementStatus[];
```

packages/client/src/clients/in-app-message/message-client.ts
thomaswhyyou: Hm that feels kinda wrong to me, that this function can return nothing or something (including how we do in the feeds client). It feels like it should return a uniform value with a network status indicating an in-flight request. 

Don't want to block on this though, we can come back to this.

thomaswhyyou (approved):
Overall I think this is a good starting point and something we can land then iterate more!

packages/client/src/clients/in-app-message/message-client.ts
connorlindsey: Yep!

packages/client/src/clients/in-app-message/message-client.ts
connorlindsey: Yeah, I'll leave a todo comment. I wouldn't be opposed to including the network status in the response (although you can grab that from the store for the given query), but I'd want to think through what the status should be

packages/client/src/clients/in-app-message/message-client.ts
thomaswhyyou: Actually just saw that we already have on in the next PR 😄

packages/client/src/clients/in-app-message/index.ts
cjbell, line 1: I think the other "clients" here are pluralized

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export interface InAppMessageResponse {
+  items: InAppMessage[];
+  pageInfo: PageInfo;
+}
+
+export interface InAppMessagesQueryInfo {
+  networkStatus: NetworkStatus;
+  loading: boolean;
+  data?: InAppMessageResponse;
```
cjbell, line 64: Do we need to keep this per-query?

packages/client/src/clients/in-app-message/types.ts
```diff
@@ -0,0 +1,103 @@
...
+  link_clicked_at: string | null;
+  source: NotificationSource;
+}
+
+export interface InAppMessageResponse {
+  items: InAppMessage[];
+  pageInfo: PageInfo;
+}
+
+export interface InAppMessagesQueryInfo {
```
cjbell, line 61: A big q I have is if this state needs to be hoisted here, or if it would be better served at the message-client level. What's your thinking here about why this needs to be shared?

packages/client/src/clients/in-app-message/channel-client.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Manages the configuration for an in app channel.
+ * Stores all fetched messages to support optimistic updates.
+ */
+export class InAppChannelClient {
+  public store: InAppStore;
+
+  constructor(
+    readonly knock: Knock,
+    readonly channelId: string,
```
cjbell, line 14: It feels like there's also likely other shared/default options that can be passed here that can be shared across all message clients (tenant is a good example)

packages/client/src/clients/in-app-message/message-client.ts
```diff
@@ -0,0 +1,113 @@
...
+          networkStatus: options.__loadingType ?? NetworkStatus.loading,
+        },
+      },
+    }));
+
+    const result = await this.knock.client().makeRequest({
+      method: "GET",
+      url: `/v1/users/${this.knock.userId}/in-app-messages/${this.channelClient.channelId}/${this.messageType}`,
+      params: queryParams,
+    });
```
cjbell, line 64: Would like to see this being pulled off into `user` module on the client as `getInAppMessages` or something similar

packages/client/src/clients/in-app-message/message-client.ts
```diff
@@ -0,0 +1,113 @@
...
+          messages[message.id] = message;
+          return messages;
+        }, state.messages),
+        // Store query results
+        queries: {
+          ...state.queries,
+          [queryKey]: queryInfo,
+        },
+      };
+    });
```
cjbell, line 109: I think this might be better encapsulated on the channelClient as a method that's invoked

packages/client/src/clients/in-app-message/message-client.ts
```diff
@@ -0,0 +1,113 @@
...
+      this.channelClient.store.setState((state) => ({
+        ...state,
+        queries: {
+          ...state.queries,
+          [queryKey]: {
+            ...queryState,
+            networkStatus: NetworkStatus.error,
+          },
+        },
+      }));
```
cjbell, line 76: Same here, would be nice to have this as a method on the channel client

cjbell (commented):
few qs here, generally looks good and biggest thing is really the queries on the channel client

packages/client/src/clients/in-app-message/index.ts
connorlindsey: Ah yeah. Was following feed which is singular, but makes sense to make it plural

packages/client/src/clients/in-app-message/types.ts
connorlindsey: Not the full message data, but we at least need the ids so that we can look up the messages under `store.messages`. I'll update the types / store to only store the ids per query

packages/client/src/clients/in-app-message/message-client.ts
connorlindsey: Agreed!

packages/client/src/clients/in-app-message/channel-client.ts
connorlindsey: Sounds good to me! So we'd cascade down options like ChannelClient > MessageClient > Fetch for JS. In React, that'd be KnockInAppChannelProvider > useInAppMessage(s) > messageClient.fetch (if called directly)

packages/client/src/clients/in-app-message/types.ts
connorlindsey: @cjbell Just to double check, do you mean all query info or just the duplicated data? 

I'm partial to having query data at the top level store. Having a single store is easier to work with and manage instances of. Maybe an edge case, but this also makes it so that duplicated queries update together like if you had multiple instances of a message in different parts of a page although maybe thats a stretch.

PR #227, opened 2024-10-01
feat: guide client and provider + ootb components

connorlindsey:


connorlindsey:
* **#228** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/228?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#227** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/227?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

thomaswhyyou (approved):

packages/react-core/src/modules/guide/context/KnockGuideProvider.tsx
```diff
@@ -0,0 +1,60 @@
...
+  try {
+    knock = useKnockClient();
+  } catch (_) {
+    throw new Error("KnockGuideProvider must be used within a KnockProvider");
+  }
+
+  const triggerParams = useStableOptions({ data, tenant });
+
+  const knockGuideClient = React.useMemo(() => {
+    return new KnockGuideClient(knock, triggerParams);
```
thomaswhyyou, line 40: I think this bit here is worth flagging and drawing attention for feedbacks: trigger params are a part of the guide client constructor arg, which are passed in as props to the provider and used in a dep array for memoizing the client. 

This means, if trigger params change (e.g. `data` or `tenant`) then we re-instantiate the client effectively resetting it. This is different than filter params like `message_type` that can be passed in to the fetch function in an ad hoc fashion, because trigger params affect eligibility evaluations as well as the rendered content. This way we make sure each instantiated client can have guides that are consistently evaluated and rendered.

connorlindsey (commented):
Thought this notification was a review request but just an old branch. Either way, this is looking good 😅

thomaswhyyou:
> Thought this notification was a review request but just an old branch. Either way, this is looking good 😅

@connorlindsey Ah yeah I've been working off the same old branch/PR we started 😅 

I haven't requested for review yet while I was iterating on it, but at this point all the major pieces and the core APIs are pretty set I think. I will need to clean up by removing the old in-app messaging related code, then will request review early next week.

PR #229, opened 2024-10-01
fix: single useStore in useInAppMessages, only message id on query info, etc.

connorlindsey:
- Update store usage and add helpers for getting and modifying store state
- Only store message ids in query info
- Move fetching in app messages to users class
- Add option to pass default options to channel client (and provider) and merge with default options in the message client. There are then 3 layers to set options (Channel client > Message client > Fetch)


connorlindsey:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/229?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#229** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/229?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#228** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/228?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#227** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/227?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

thomaswhyyou:
Will look at all the open PRs shortly

packages/client/src/clients/in-app-messages/types.ts
```diff
@@ -21,15 +24,21 @@ export interface InAppMessage<TContent = GenericData, TData = GenericData> {
   source: NotificationSource;
 }
 
-export interface InAppMessageResponse {
-  entries: InAppMessage[];
+export interface InAppMessageResponse<
```
thomaswhyyou, line 27: Not a blocker for this PR, but we should take a full pass later to get all naming things right/consistent (e.g. plural vs singular).

packages/react-core/src/modules/in-app-message/hooks/useInAppMessages.ts
```diff
@@ -12,13 +12,19 @@ import { useInAppChannel } from "../context";
 
 export interface UseInAppMessagesOptions extends InAppMessagesClientOptions {}
 
-export interface UseInAppMessagesResponse<TContent, TData> {
+export interface UseInAppMessagesResponse<
```
thomaswhyyou, line 15: Not for this PR, but again later we should take a full pass at removing all exports that aren't used outside and meant to be used by end users.

packages/react-core/src/modules/in-app-message/hooks/useInAppMessages.ts
```diff
@@ -29,37 +35,9 @@ export const useInAppMessages = <TContent = GenericData, TData = GenericData>(
...
+  const { messages, networkStatus, loading } = useStore(
     inAppChannelClient.store,
-    (state) => {
-      const query = state.queries[inAppMessagesClient.queryKey];
-      return {
-        networkStatus: query?.networkStatus ?? NetworkStatus.ready,
-        loading: query?.loading ?? false,
-      };
-    },
+    inAppMessagesClient.getQueryInfoSelector<TContent, TData>,
```
thomaswhyyou, line 40: I'm wondering out loud if this helper really belongs to `inAppMessagesClient` i.e. is it really `inAppMessagesClient` concern? 

Asking because from what I can tell, this selector returns an object that is relevant to the `useInAppMessage` hook, and not something you care if you are using `inAppMessagesClient` hook directly without this hook? If so, then this selector should probably live in this file?

packages/client/src/clients/in-app-messages/message-client.ts
```diff
@@ -69,75 +73,89 @@ export class InAppMessagesClient {
...
-      },
-    }));
-
-    // TODO: Move to method on user.getInAppMessages
-    const result = await this.knock.client().makeRequest({
-      method: "GET",
-      url: `/v1/users/${this.knock.userId}/in-app-messages/${this.channelClient.channelId}/${this.messageType}`,
-      params: queryParams,
+    this.channelClient.setQueryState(this.queryKey, {
+      ...queryState,
```
thomaswhyyou, line 78: I feel like this helper should be made into one that updates a query status only. Which means, 
* `...queryState` part should be done as part of the helper
* the helper to be renamed to something like `setQueryStatus`

packages/client/src/clients/in-app-messages/message-client.ts
```diff
@@ -69,75 +73,89 @@ export class InAppMessagesClient {
...
+          pageInfo: response.pageInfo,
+        },
       };
-    }
 
-    const response: InAppMessageResponse = {
-      entries: result.body.entries,
-      pageInfo: result.body.page_info,
-    };
+      this.channelClient.store.setState((state) => {
```
thomaswhyyou, line 99: And I think this can be also abstracted to a helper called `setQueryResult` or `setQueryResp` that:
* Takes the query key and the response
* Builds a query info and updates both `messages` and `queries`

So you end up with a set of two helpers: `setQueryStatus` and `setQueryResult`, of which the latter is used when we have a successful response. Wdyt?

thomaswhyyou (commented):
Left a few small comments/questions.

packages/react-core/src/modules/in-app-message/hooks/useInAppMessages.ts
connorlindsey: Ack, got a todo for that

packages/client/src/clients/in-app-messages/types.ts
connorlindsey: Ack, got a todo for that

packages/react-core/src/modules/in-app-message/hooks/useInAppMessages.ts
connorlindsey: I placed it in the message client so that it can be used outside of React. If someone were to use the message client with say Vue, they would need to use this same selector to get query info for the given message + filters. We can't fully abstract the fact that we're using Tanstack store because for the store to be reactive, it needs to be accessed using the correct `useStore` hook for the given framework. Given that, I think providing helper methods like this is a good solution so that non-React users only have to do something like:

```
// my-component.vue
<script setup>
import { useStore } from '@tanstack/vue-store';
import { store } from './store';

const { ...message info } = useStore(store, (state) => inAppMessageClient.getQueryInfoSelector(store))
</script>
```

packages/client/src/clients/in-app-messages/message-client.ts
connorlindsey: Yeah, I'm good with separating this into two more specific helpers

packages/react-core/src/modules/in-app-message/hooks/useInAppMessages.ts
thomaswhyyou: Ah okay I see. 

Just to double check, this query info data is only set by the `fetch()` call, right? Meaning, the "query"  in `getQueryInfoSelector()` helper refers to `fetch()`, and the selector would return empty data if you haven't called `fetch()`?

connorlindsey: Yeah, it's only set by fetch currently. The realtime connection will also update the same info when a new message is received (e.g. the list of messages associated with a set of filters/message type)

thomaswhyyou (approved):
👍

PR #232, opened 2024-10-03
feature: add addl slack scopes and ex.

JEverhart383:


examples/slack-kit-example/app/actions/fetch-user-id.ts
```diff
@@ -0,0 +1,63 @@
...
+  const response = await fetch(
+    `https://slack.com/api/users.lookupByEmail?email=${email}`,
+    {
+      headers: {
+        Authorization: `Bearer ${channelData.data.token.access_token}`,
+      },
+    },
+  );
+  const data = await response.json();
+  console.log("data", data);
```
connorlindsey, line 22: Do you want to keep these console logs in? If so, maybe worth formatting them a bit more, like by making the text more detailed (e.g. `console.log("[Knock] User data:", data);`

JEverhart383: Good catch, removed the ones that weren't error related

PR #243, opened 2024-10-11
feat(kno-6559): set up @knocklabs/expo package

mattmikolay:
# Summary

This PR sets up a new package `@knocklabs/expo` for use with React Native + Expo projects, and moves the existing `KnockExpoPushNotificationProvider` plus all required Expo dependencies to `@knocklabs/expo`.

# Background

At the moment, `@knocklabs/react-native`’s `package.json` includes all Expo-related dependencies as optional peer dependencies:

https://github.com/knocklabs/javascript/blob/f902b998fdcc15161f6eb2f3549e9a14a4d861ed/packages/react-native/package.json#L43-L64

However, `KnockExpoPushNotificationProvider.tsx` imports these dependencies using static imports:

https://github.com/knocklabs/javascript/blob/f902b998fdcc15161f6eb2f3549e9a14a4d861ed/packages/react-native/src/modules/push/KnockExpoPushNotificationProvider.tsx#L7-L9

This causes build errors when Metro (the bundler used by React Native) bundles an application that depends on `@knocklabs/react-native` but does _not_ include Expo dependencies.

It’s possible to refactor `KnockExpoPushNotificationProvider` to import Expo modules dynamically using [`require`](https://metrobundler.dev/docs/module-api/#require). Support for loading optional dependencies via `require` was added to [Metro v0.59.0](https://github.com/facebook/metro/releases/tag/v0.59.0) by facebook/metro#511.

However, optional dependencies in React Native are subject to some longstanding issues, e.g. facebook/metro#836 and facebook/metro#1256. (And sure enough, during my testing locally, I encountered these issues.) Even if we load Expo modules using `require` and avoid these bugs, it’s highly likely future refactoring of the code would cause the build errors to occur again. For this reason, we feel it’s safer to migrate all code in `@knocklabs/react-native` that relies upon Expo to a new Expo-specific SDK.

# Testing

I tested everything locally and confirmed the following:
- Both `@knocklabs/react-native` and `@knocklabs/expo` successfully receive notifications via `KnockFeedProvider`
- Push notifications are successfully received via the `KnockExpoPushNotificationProvider` after moving it to the `@knocklabs/expo` package
- The build issues previously caused by Expo module imports in `@knocklabs/react-native` have been fixed


packages/expo/package.json
```diff
@@ -0,0 +1,80 @@
+{
+  "name": "@knocklabs/expo",
+  "version": "0.3.1",
```
cjbell, line 3: Just to confirm, is this right for the initial version? Do we not want to start at 0.1.0?

cjbell (commented):
one q

packages/expo/package.json
mattmikolay: We can definitely start at v0.1.0. I wasn’t sure what’s preferable: restarting version numbering at 0 since this is a new package, or starting with the same version number as our RN SDK since the new package is essentially a fork.

cjbell (approved):

PR #252, opened 2024-10-15
feat(kno-6559): add KnockPushNotificationProvider

mattmikolay:
Now that #243 moved `KnockExpoPushNotificationProvider` to the new `@knocklabs/expo` package, we need to provide a way for non-Expo React Native applications to register push tokens to a channel for the current user.

This PR adds a new context provider `KnockPushNotificationProvider` to `@knocklabs/react-native`. This context provider exposes two functions, `registerPushTokenToChannel` and `unregisterPushTokenFromChannel`. These can be used to manually register/unregister push tokens retrieved from a provider such as Firebase Cloud Messaging.

Additionally, `KnockExpoPushNotificationProvider` has been refactored to use this new `KnockPushNotificationProvider`.

I tested these changes on an Android device and confirmed push notifications can still be received using the Expo sample app.


packages/react-native/src/modules/push/KnockPushNotificationProvider.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+const KnockPushNotificationContext = createContext<
+  KnockPushNotificationContextType | undefined
+>(undefined);
+
+export interface KnockPushNotificationProviderProps {
+  children?: React.ReactElement;
+}
+
+export const KnockPushNotificationProvider: React.FC<
```
mattmikolay, line 21: For what it’s worth, this `KnockPushNotificationProvider` does not use any APIs specific to React Native. Alternatively, we could add this to `@knocklabs/react-core`.

`@knocklabs/react-native` already re-exports everything from `@knocklabs/react-core`, so it would still be available for use via our React Native SDK:

https://github.com/knocklabs/javascript/blob/5d60ed0815a7697bc9a2e59c0ba5b71914e04084/packages/react-native/src/index.ts#L2

cjbell (approved):
💯

PR #279, opened 2024-11-08
fix: refetch in-app messages when query options change 

thomaswhyyou:
Fixes this issue: https://knocklabs.slack.com/archives/C079G6X3DD4/p1731085160770939

* Need to re-run the effect upon getting new options, so adding inAppMessagesClient to the useEffect deps array in the in-app message client
* Uses `fast-deep-equal` instead of `zustand/shallow` since the latter does not handle array values, and some of our query params can be arrays.

### DIFF
--- a/packages/react-core/src/modules/core/hooks/useStableOptions.ts
+++ b/packages/react-core/src/modules/core/hooks/useStableOptions.ts
@@ -1,5 +1,5 @@
+import fastDeepEqual from "fast-deep-equal";
 import { useMemo, useRef } from "react";
-import shallow from "zustand/shallow";
 
 export default function useStableOptions<T>(options: T): T {
   const optionsRef = useRef<T>();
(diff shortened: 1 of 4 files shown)


thomaswhyyou:
* **#279** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/279?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `next`

<h2></h2>

 Join @thomaswhyyou and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

connorlindsey (approved):

packages/react-core/src/modules/core/hooks/useStableOptions.ts
```diff
@@ -1,13 +1,13 @@
+import fastDeepEqual from "fast-deep-equal";
```
thomaswhyyou, line 1: I considered rolling something manually on our own, but decided against it because there's always a risk of a leaky implementation and no guarantee it'd be faster or more performant than existing solutions. 

I did some google search and it seems like this one is what people seem to recommend as the fastest option. I checked the bundle size and it clocks in at ~660b compressed. 

Unclear how this compares to `zustand/shallow` performance wise though, but I suspect it's comparable assuming objects that are not huge or deeply nested.

CleanShot 2024-11-08 at 15 51 04@2x

cjbell (approved):

PR #341, opened 2025-01-08
chore: render body in ootb components as inner html

thomaswhyyou:
### What changed?
Following the update that converted the body fields of OOTB message types to markdown (https://github.com/knocklabs/control/pull/4362), this PR now renders the rendered body/html content as an inner html in the components. This should handle the rendered content whether it was a text or markdown field.

### Tasks
[KNO-7559](https://linear.app/knock/issue/KNO-7559/[sdk]-change-ootb-message-type-body-fields-to-markdown)

### Screenshots or Loom

CleanShot 2025-01-08 at 10 36 55@2x

### DIFF
--- a/packages/react/src/modules/in-app-messages/components/Banner/Banner.tsx
+++ b/packages/react/src/modules/in-app-messages/components/Banner/Banner.tsx
@@ -70,9 +70,11 @@ const Body: React.FC<{ body: string } & React.ComponentPropsWithRef<"div">> = ({
   ...props
 }) => {
   return (
-    <div className={clsx("knock-iam-banner__body", className)} {...props}>
-      {body}
-    </div>
+    <div
+      className={clsx("knock-iam-banner__body", className)}
+      dangerouslySetInnerHTML={{ __html: body }}
+      {...props}
+    />
   );
 };
 Body.displayName = "BannerView.Body";
--- a/packages/react/src/modules/in-app-messages/components/Banner/styles.css
+++ b/packages/react/src/modules/in-app-messages/components/Banner/styles.css
@@ -24,6 +24,12 @@
   font-weight: var(--knock-weight-regular);
   line-height: var(--knock-leading-3);
 }
+.knock-iam-banner__body > :first-child {
+  margin-top: 0;
+}
+.knock-iam-banner__body > :last-child {
+  margin-bottom: 0;
+}
 .knock-iam-banner__actions {
   display: flex;
   align-items: center;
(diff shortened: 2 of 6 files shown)


thomaswhyyou:
* **#341** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/341?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/341?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `next`


<h2></h2>

packages/react/src/modules/in-app-messages/components/Banner/styles.css
```diff
@@ -24,6 +24,12 @@
   font-weight: var(--knock-weight-regular);
   line-height: var(--knock-leading-3);
 }
+.knock-iam-banner__body > :first-child {
```
thomaswhyyou, line 27: Removing vertical margins at the top and the bottom to match our design, since elements like `<p>` tags usually have some margins by default.

packages/react/src/modules/in-app-messages/components/Banner/Banner.tsx
```diff
@@ -70,9 +70,11 @@ const Body: React.FC<{ body: string } & React.ComponentPropsWithRef<"div">> = ({
   ...props
 }) => {
   return (
-    <div className={clsx("knock-iam-banner__body", className)} {...props}>
-      {body}
-    </div>
+    <div
+      className={clsx("knock-iam-banner__body", className)}
+      dangerouslySetInnerHTML={{ __html: body }}
```
connorlindsey, line 75: Agreed this works well for the OOTB schemas that will now be markdown. Are there any drawbacks though in terms of being able to reuse these components for custom schemas? Wondering if it'd be worth it to provide an escape mechanism to still render e.g. `body` as a regular child component (or keep that as the regular and only render using inner html for the 'default' component that is tied to the OOTB schemas). I'd lean towards not worrying about it at the moment.

connorlindsey (approved):

packages/react/src/modules/in-app-messages/components/Banner/Banner.tsx
thomaswhyyou: > Are there any drawbacks though in terms of being able to reuse these components for custom schemas?

My thinking was that you would use these components if you are using the ootb message type with the provided schemas provided as is. You can customize rendering side of the ootb message types in various ways, including using your own component(s). But if you have a custom schema then you should bring your own component that matches the schema.

PR #348, opened 2025-01-11
chore: make react* sdks react 19 compatible

thomaswhyyou:
This PR includes changes aimed to add support for React 19 with the `react` and `react-core` packages.

The first immediate thing is there are several external dependencies in `react` and `react-core` that need be updated to the latest versions so they support or allow react 19 as a peer dependency. 

There was one exception to this which was `react-popper` - the [react-popper](https://github.com/floating-ui/react-popper) project was deprecated earlier last year in favor of a new project [floating-ui](https://floating-ui.com/) as a successor. The last react-popper version has react 18 as a peer dependency so we need to migrate to the new floating-ui lib, but the API has changed slightly so it involved some code changes. 

There are a few remaining questions/issues:
* We will most likely need to remove `react-popper-tooltip` from `react` for the same reason explained above. `react-popper-tooltip` itself [accepts](https://github.com/mohsinulhaq/react-popper-tooltip/blob/master/package.json#L55-L56) react 19, but it has a dependency to [react-popper 2.3.0](https://github.com/mohsinulhaq/react-popper-tooltip/blob/master/package.json#L61), which has a deep dep up to react 18.
* Upgrading radix to the latest versions may cause issues for customers running react 16, based one this previous PR: https://github.com/knocklabs/javascript/pull/273
* There _may_ be issues with [zustand](https://github.com/knocklabs/javascript/blob/thomas-kno-7523-js-support-react-19-in-react-sdks-2/packages/react-core/package.json#L55) and react 19. I tried setting `react` and `@types/react` to 19 in our `react-core` package, and the type check was failing and a couple of the errors were coming from zustand func calls. My first thought was to upgrade `zustand` to the latest and make any necessary changes, but I realized we are quite behind and I vaguely remember us having issues upgrading `zustand`?

I think the ideal solution is to resolve the issues above, but still keep react and react-dom deps for dev in our packages still at 18 because it seems like our monorepo setup forces a single consistent version across all of our packages and example apps, and upgrading them may be too big of a lift.

### DIFF
--- a/examples/nextjs-example/package.json
+++ b/examples/nextjs-example/package.json
@@ -23,7 +23,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-icons": "^5.4.0",
-    "swr": "^2.2.5",
+    "swr": "^2.3.0",
     "uuid": "^11.0.4"
   },
   "devDependencies": {
--- a/packages/react-core/package.json
+++ b/packages/react-core/package.json
@@ -46,12 +46,12 @@
     "url": "https://github.com/knocklabs/javascript/issues"
   },
   "peerDependencies": {
-    "react": "^16.11.0 || ^17.0.0 || ^18.0.0"
+    "react": "^16.11.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
   },
   "dependencies": {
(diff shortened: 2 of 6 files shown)


MikeCarbone:
Closing in favor of #419

Issue #317, opened 2024-12-12
Support React 19


danilofuchs:
Knock react does not declare React 19 as a peer dependency, and uses old versions of Radix

![image](https://github.com/user-attachments/assets/c3370d9a-e03d-4c4e-8e4a-99f9d76405cc)
![image](https://github.com/user-attachments/assets/f74d3212-3fea-4a33-b43b-e10faf2bd90b)

jdhiro:
+1, this is the only package I still have to override deps on.

aldenquimby:
Looks like radix just added react 19 support: https://github.com/radix-ui/primitives/pull/3229

pippinmole:
Draft PR is open: https://github.com/knocklabs/javascript/pull/348

MikeCarbone:
Actively working on this, stay tuned 🙂

jdhiro:
Sweet, upgrading!

MikeCarbone:
@jdhiro Awesome! Let me know if you have any issues. By the way, I released a new RC version to fix a small, unintentional breaking change. I updated my comment above to show the latest version to use (`rc.1.0` as of today)

MikeCarbone:
We released the third release candidate last night (3.0) — if you feel inclined, please check it out! Let me know if you encounter any issues (and would also love to know if you _don't_ encounter any issues!)

jdhiro:
Thanks @MikeCarbone !

We're active users and big fans of Knock. We'll use the new RC in our next deployment. Cheers! 🍻

MikeCarbone:
This is now complete with the merge of #448 ! Going to close this issue.

Here is the changelog from our website: [https://knock.app/changelog/2025-03-13](https://knock.app/changelog/2025-03-13)

You can now update to the latest versions (non-breaking):


- [@knocklabs/client: v0.12.0](https://github.com/knocklabs/javascript/releases/tag/%40knocklabs%2Fclient%400.12.0)
- [@knocklabs/react: v0.6.0](https://github.com/knocklabs/javascript/releases/tag/%40knocklabs%2Freact%400.6.0)
- [@knocklabs/react-core: v0.5.0](https://github.com/knocklabs/javascript/releases/tag/%40knocklabs%2Freact-core%400.5.0)

Supporting React 19 in our React Native & Expo packages is on our radar. We'll circle back to those packages when other dependencies catch up.

Since the underlying dependency on `@knocklabs/client` was updated, we released a new version for these packages as well.

[...]

PR #419, opened 2025-02-20
[JS] Support React 19 in React SDKs

MikeCarbone:
# [JS] Support React 19 in React SDKs

## **Breaking changes:**
- By upgrading Zustand to v4, `feed.store` now uses the `UseBoundStore<StoreApi<FeedStoreState>>` type (previously `StoreApi<FeedStoreState>`), meaning users no longer need their own implementation of `zustand` to use the store directly.

Users who relied only on `useNotificationStore` to access the store will not be affected, only users who use `feedClient.store` directly.

---
### Enables support for React 19

This PR adds support for React 19 by including it within the `peerDependencies` of the various packages.

### Removes `react-popper` dependency

This dependency was deprecated. I opted to use our existing dependency `@popperjs/core` for popup elements

### Upgrades `zustand` to v4

Introduces a breaking change when users use the `feedClient.store` directly.

- [Upgrade guide to v4](https://zustand.docs.pmnd.rs/migrations/migrating-to-v4)

We decided not to upgrade to `v5` because that would drop support for React 16, which we should support.

### Upgrades internal `Telegraph` components for React 19 support

In order for these packages to support React 19, we had to first [enable our Telegraph components to support React 19](https://github.com/knocklabs/telegraph/pull/409). After that was done, I bumped all of the Telegraph dependencies in this PR.

### Next.js Example Now Uses Telegraph Components

Now that we have the `@telegraph` design system supporting React 19, we were able to use it in the `next-js-example` app.

**Before**         |  **After**
:-------------------------:|:-------------------------:
Screenshot 2025-02-21 at 2 15 58 PM  |  Screenshot 2025-02-21 at 2 16 34 PM

---
### A note on `react-native` package:

As of 2 days ago, [React 19 is now available on React Native](https://reactnative.dev/blog/2025/02/19/react-native-0.78), however, dependent packages like [react-shallow-renderer](https://www.npmjs.com/package/react-shallow-renderer) need time to catch up and support this. So for now, we will leave the `react-native` package as is. This causes a dependency warning when running `yarn install` at the root of this repository, but that won't affect individual builds.

---

[Linear issue](https://linear.app/knock/issue/KNO-7523/[js]-support-react-19-in-react-sdks)

[Addresses issue "Support React 19 #317"](https://github.com/knocklabs/javascript/issues/317)


MikeCarbone:
* **#419** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/419?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/419?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/react/src/modules/feed/components/NotificationFeed/MarkAsRead.tsx
```diff
@@ -14,8 +15,10 @@ export const MarkAsRead: React.FC<MarkAsReadProps> = ({ onClick }) => {
   const { useFeedStore, feedClient, colorMode } = useKnockFeed();
   const { t } = useTranslations();
 
-  const unreadItems = useFeedStore((state) =>
-    state.items.filter((item) => !item.read_at),
+  // Without useShallow, the component will get itself in a re-render loop
+  // https://zustand.docs.pmnd.rs/migrations/migrating-to-v5#requiring-stable-selector-outputs
+  const unreadItems = useFeedStore(
+    useShallow((state) => state.items.filter((item) => !item.read_at)),
   );
```
cjbell, line 22: Could we refactor this to be:

```tsx
const {items} = useState(state => state.items);
const unreadItems = items.filter((item) => !item.read_at);
```

Maybe that needs to be in a useMemo for the filter? Not sure.

MikeCarbone:
Finally, I was able to get these packages running locally within a project running React 19. However it took a bit of tinkering to get all of the dependencies to resolve to React 19 correctly. Here's what happened...

## Problem Running Within Example Apps

I was not able to get the `nextjs-example` to run using React 19 within the repo. The project would not build because of conflicting React versions. I thought that since `react` and `react-dom` were listed in `devDependencies` _and_ `peerDependencies` within our `react` and `react-core` packages, that these packages would lean on the local version of React (v19 in this case). But it seems packages would only do this until a certain point. No matter what I tried, our `react` package wanted its own copy of v18 React.

### The Errors

Each time, the project would either
a) fail to build and try to revert back the project to use React v18
b) force a 19 resolution, try to use both, and fail to build, creating issues like `dispatcher is null`

### Environment?

This _could be__ because of the environment. Using these local packages might lead to some jank within the hoisted `node_modules` setup causing conflicts. I tried to prevent hoisting `node_modules` between packages and examples but the issue still persisted within the individual example app.

### Simulating a Fresh Install

So, to test, I moved everything to a new nextjs project and moved the packages over by hand, only pulling the package.json and `dist` folders after build, and installed them as local packages. Presumably, this would mirror the setup if we pushed this live.

The issues persisted. Our React packages insisted on using their own copy of React 18, despite having a peerDependency of React 19 and only including `react` as a devDependency.

### Solution

I was able to get everything running by removing `react` and `react-dom` from the `devDependencies` of the `packages/react` and `packages/react-core` packages (and keeping those deps within `peerDependencies`). This allowed those packages to successfully rely on the project's version of React 19.

## Open Questions

- Would everything work fine if we just distributed these packages remotely? Is this because of the local environments?
- If no to the question above, why would a `devDependency` of React 18, with a `peerDependency` of React 19 not allow us use the local project's React 19 version? Why does our package _insist_ on using its own React 18? We do this in our @telegraph components and it seems to work fine.
- If above is a real issue, will removing `react` and `react-dom` from our dev dependencies cause issues? Presumably, if a user is installing Knock's `react` package, they will have a copy of react installed. Our package should be able to rely on this as long as it's a version within our peerDependencies. But will this cause issues with development locally?

examples/nextjs-example/pages/headless.tsx
```diff
@@ -41,26 +42,31 @@ const HeadlessFeed = ({
     { tenant },
   );
 
-  const { items, metadata } = useNotificationStore(feedClient);
+  // Example of using a selector to access a subset of the store state.
+  const { items, metadata } = useNotificationStore(feedClient, (state) => ({
+    items: state.items,
+    metadata: state.metadata,
+  }));
```
cjbell, line 49: Does this always need a selector now?

examples/nextjs-example/styles/globals.css
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    100% {
+        transform: rotate(360deg);
+    }
+}
+
+a {
+    text-decoration: none;
+}
```
cjbell, line 27: Spacing in this file is weird

cjbell (commented):
few notes!

examples/nextjs-example/pages/headless.tsx
MikeCarbone: Nope this was just an example usage. I'll add a comment specifying that!

packages/react/src/modules/feed/components/NotificationFeed/MarkAsRead.tsx
MikeCarbone: the `useShallow` approach looks to be the proper way to access computed property based on state with the new Zustand: [https://zustand.docs.pmnd.rs/guides/prevent-rerenders-with-use-shallow](https://zustand.docs.pmnd.rs/guides/prevent-rerenders-with-use-shallow)

still want me to try `useMemo`?

packages/react/src/modules/feed/components/NotificationCell/ArchiveButton.tsx
```diff
@@ -1,7 +1,7 @@
 import { FeedItem } from "@knocklabs/client";
 import { useKnockFeed, useTranslations } from "@knocklabs/react-core";
-import React, { MouseEvent, useCallback } from "react";
-import { usePopperTooltip } from "react-popper-tooltip";
+import { createPopper } from "@popperjs/core";
```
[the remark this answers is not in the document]
MikeCarbone: Tried this -- introduces too many changes (with at least one breaking change) to our `NotificationFeedPopover`. Out of scope for this right now

cjbell (approved):
let's goooooooooooooooooo

.changeset/red-owls-jog.md
```diff
@@ -0,0 +1,12 @@
+---
```
connorlindsey, line 1: Should this be at least a minor release given there are breaking changes?

examples/nextjs-example/pages/index.tsx
```diff
@@ -49,86 +51,98 @@ export default function Home() {
         defaultFeedOptions={{ tenant }}
       >
         <NotificationFeedContainer>
-          <Box maxW="520px" mx="auto" py={12}>
-            <Flex mb={8}>
-              <Box>
-                <Heading size="md" mb={2}>
+          <Box style={{ maxWidth: "500px" }} mx="auto" py="12">
```
connorlindsey, line 54: Nit: Ideally we use telegraph constants wherever possible. In the case of a demo app like this, I think just setting a reasonable pixel value is totally okay but wanted to still call it out since having to use `style` with telegraph is often a smell.

packages/client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@knocklabs/client",
-  "version": "0.11.4",
+  "version": "0.12.0",
```
connorlindsey, line 3: I don't think you need to manually bump the version. That should be handled by changesets.

packages/client/upgrade-guides/to-v0.12.0.md
```diff
@@ -0,0 +1,123 @@
...
+This update impacts the following packages:
+- Upgrading `@knocklabs/client` to v0.12.0
+- Upgrading `@knocklabs/react-core` to v0.5.0
+- Upgrading `@knocklabs/react` to v0.6.0
+- Upgrading `@knocklabs/react-native` to v0.6.0
+- Upgrading `@knocklabs/expo` to v0.3.0
+
+## Breaking changes
+
+As part of this update, we've upgraded our dependency on `zustand` to use the latest version. This means that we've had to make some breaking changes to the API around accessing the data store.
```
connorlindsey, line 14: Nit: Could call out the specific version so users can look it up if needed.

packages/client/upgrade-guides/to-v0.12.0.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Changes to `feedClient.store`
+
+Consumers no longer need to use their own Zustand dependency to access the store. You can now access the store directly using the `store` property on the `feedClient` object.
```
connorlindsey: Nit: I think we should stay consistent using `you` when referring to the reader rather than referring to them as `Consumers` then swapping to directly speaking to them.

packages/expo/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@knocklabs/expo",
-  "version": "0.2.10",
+  "version": "0.3.0",
```
connorlindsey, line 3: Same to not needing to edit this directly.

packages/react-core/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@knocklabs/react-core",
   "description": "A set of React components to build notification experiences powered by Knock",
   "author": "@knocklabs",
-  "version": "0.4.0",
+  "version": "0.5.0",
```
connorlindsey, line 5: Same

packages/react-core/package.json
```diff
@@ -46,16 +46,17 @@
     "url": "https://github.com/knocklabs/javascript/issues"
   },
   "peerDependencies": {
-    "react": "^16.11.0 || ^17.0.0 || ^18.0.0"
+    "react": "^16.11.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
```
connorlindsey, line 49: Are there any issues with backwards compatibility to e.g. `16.11.0` with the changes we've made like Zustand? Here are the peer deps for zustand `5.0.3`

```
  "peerDependencies": {
    "@types/react": ">=18.0.0",
    "immer": ">=9.0.6",
    "react": ">=18.0.0",
    "use-sync-external-store": ">=1.2.0"
  },
  ```

connorlindsey (commented):
Have we considered dropping zustand for tanstack-store after using it for in-app-messages? I'm not strongly for it given how many changes we're already having to make here but maybe worth exploring.

packages/react-core/package.json
MikeCarbone: Aghhhh great call. Let me look deeper into this...

MikeCarbone: I downgraded to zustand 4.5.6 to preserve React 16 support 😓

examples/nextjs-example/pages/index.tsx
MikeCarbone: I wanted to match the current example as best as possible and telegraph didn't over a suitable `maxWidth` to use here. I tried!

.changeset/red-owls-jog.md
MikeCarbone: Yes! definitely

MikeCarbone:
@connorlindsey Re: tanstack-store. Chris and I talked about it but we figured it was out of scope for now. I'm a big tanstack fan, the idea sounds nice! But let's revisit that another time

PR #428, opened 2025-02-24
fix(kno-7843): fixes `trigger_data` params functionality and types

MikeCarbone:
# fix(kno-7843): consistent `trigger_data` params functionality

_Note: this PR is a branch off the React 19 updates branch. I wanted to make sure the comparison logic I changed would work with the new zustand updates_

---
This PR addresses [KNO-7843](https://linear.app/knock/issue/KNO-7843/[js-sdk]-trigger-data-query-params-dont-work-correctly-with-boolean), an issue with the SDK where specifying a `trigger_data` field when using `useNotifications()` would require stringifying booleans, and would create an awkward typing situation. It didn't work out of the box.

[As specified by our API](https://docs.knock.app/reference#get-feed), `trigger_data` field should be a string. Our SDK's typing says it should be an object, which is correct for a nice DX. This PR remediates those differences elegantly so that our usage is consistent, and if a user (like the one Matt highlighted in the issue) did an override, their code won't break on them and will respect that.

---

### Example 1: supports existing overrides. Users might have done this to get things working and we don't want that to break
```ts
const feedClient = useNotifications(
  knockClient,
  process.env.NEXT_PUBLIC_KNOCK_FEED_CHANNEL_ID!,
  {
    tenant,
    // @ts-expect-error need to send a string to the API
    trigger_data: JSON.stringify({ isEnterprise: true }),
  },
);
```

### Example 2: supports proper JS object usage, as expected
```ts
const feedClient = useNotifications(
  knockClient,
  process.env.NEXT_PUBLIC_KNOCK_FEED_CHANNEL_ID!,
  {
    tenant,
    trigger_data: { isEnterprise: true },
  },
);
```

### Example 3: supports stringified keys
```ts
const feedClient = useNotifications(
  knockClient,
  process.env.NEXT_PUBLIC_KNOCK_FEED_CHANNEL_ID!,
  {
    tenant,
    trigger_data: { "isEnterprise": true },
  },
);
```

### Example 4: and of course, supports no trigger data usage at all
```ts
const feedClient = useNotifications(
  knockClient,
  process.env.NEXT_PUBLIC_KNOCK_FEED_CHANNEL_ID!,
  { tenant },
);
```

### Example 5: Tightening `trigger_data` type (errors)

The API won't accept a nested object when passed through the `trigger_data` field (`422 invalid params`), so I tightened up that type to error for users when they try to do that.

Example error:

Screenshot 2025-02-24 at 6 43 46 PM

[API docs on trigger data](https://docs.knock.app/reference#trigger-data-filtering)

## Note: Adds `deep-equal` dependency

When using a nested key structure like the examples above, zustand's `shallow` function within `useStableOptions` was no longer doing what we needed it to do. The shallow copy would not safely recognize the nested key structure, causing an infinite re-render loop. The `deep-equal` usage fixes this. I'm not sure how this would work before!

---

[Linear ticket](mike-kno-7843-js-sdk-trigger_data-query-params-dont-work-correctly-with)
[Loom example of the fix](https://www.loom.com/share/4f63d30aa4584a5e8acf4d2340591e27?sid=d4f3d987-57c9-41d0-932c-29de38e00cf0)


MikeCarbone:
closed in favor of #479

PR #439, opened 2025-03-05
fix(kno-7523): fixes bad returns from useNotificationStore

MikeCarbone:
# fix(kno-7523): fixes bad returns from useNotificationStore

[Customers were running into issues using the new release candidate.](https://knocklabs.slack.com/archives/C085BK3QYSX/p1741195268891259) This attempts to align the functionality and typing to its original behavior.

The types are a little sketchy but I could not get it to behave the way I wanted it to without type casting. If you have suggestions please let me know!

I tested this in our `client` and `nextjs-example` repos, both with and without a selector. It appears to function great and produce accurate types.

[Here is the current `useNotificationStore` code in main for reference](https://github.com/knocklabs/javascript/blob/main/packages/react-core/src/modules/feed/hooks/useNotificationStore.ts)

Also added a new page in our nextjs example that mimics [the guide from our docs](https://docs.knock.app/in-app-ui/react/custom-notifications-ui) exactly (very similar to pages/headless but still a little different).

### DIFF
--- a/examples/nextjs-example/pages/custom.tsx
+++ b/examples/nextjs-example/pages/custom.tsx
@@ -1,13 +1,13 @@
+import { MarkdownContentBlock } from "@knocklabs/client";
 import {
   KnockProvider,
   useKnockClient,
-  useNotifications,
   useNotificationStore,
+  useNotifications,
 } from "@knocklabs/react";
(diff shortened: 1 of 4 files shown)


MikeCarbone:
Linking in #317

PR #446, opened 2025-03-11
fix(KNO-8072): set revalidateFirstPage option of useSWRInfinite to false

mattmikolay:
This PR fixes an issue with the two hooks in our React SDK that use `useSWRInfinite` internally, `useMsTeamsTeams` and `useSlackChannels`.

Both of these hooks call `setSize` within a `useEffect` so that they continuously load pages of data until the max number of teams (or channels) has been loaded, or there are no more pages available:

https://github.com/knocklabs/javascript/blob/7d68b259200bf828198bc2002bd255812d41ce12/packages/react-core/src/modules/ms-teams/hooks/useMsTeamsTeams.ts#L90-L92

Calling `setSize` to fetch the next page seems to trigger SWR’s revalidation behavior. Because [`useSWRInfinite`’s `revalidateFirstPage` option](https://swr.vercel.app/docs/pagination#parameters) is `true` by default, this means the first page is refetched every time we call `setSize`.

To fix this, we can simply set `revalidateFirstPage` to `false`.

## Videos

Tested locally with a page size of 4:

### Before

Notice there are five requests to load Slack channels, three of which load the first page:

https://github.com/user-attachments/assets/bfe68832-2930-4f20-b518-9d61001cece2

### After

https://github.com/user-attachments/assets/2132e5c2-9f1c-4c6b-a356-665749e2638c

### DIFF
--- a/packages/react-core/src/modules/ms-teams/hooks/useMsTeamsTeams.ts
+++ b/packages/react-core/src/modules/ms-teams/hooks/useMsTeamsTeams.ts
@@ -10,7 +10,7 @@ const MAX_COUNT = 1000;
 
 const QUERY_KEY = "MS_TEAMS_TEAMS";
 
-type UseMsTeamsTeamsProps = {
+type UseMsTeamsTeamsOptions = {
   queryOptions?: MsTeamsTeamQueryOptions;
 };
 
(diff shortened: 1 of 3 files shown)


packages/react-core/src/modules/ms-teams/hooks/useMsTeamsTeams.ts
```diff
@@ -10,7 +10,7 @@ const MAX_COUNT = 1000;
 
 const QUERY_KEY = "MS_TEAMS_TEAMS";
 
-type UseMsTeamsTeamsProps = {
+type UseMsTeamsTeamsOptions = {
```
mattmikolay, line 13: I fixed this naming. Hooks don’t have props. 🪝

connorlindsey (approved):
Nice fix! Really appreciate the video showing the network requests 👌

PR #457, opened 2025-03-20
feat(8086): add support for date range filters in the feed

meryldakin:
This adds support for filtering messages by a given date range.

It also removes `optimizeDeps` from the client example app which caused the dependency to be cached.


MikeCarbone (approved):
looks gooooood

PR #495, opened 2025-04-15
fix: useNotifications memory leak

kylemcd:
### Description
Refactors `useNotifications` to use `useEffect` instead of `useMemo` for initializing the Knock feed client. Previously, `useMemo` was used to set up the feed and call `dispose()` when dependencies changed. However, `useMemo` is not designed for side effects or cleanup—it simply memoizes a return value and does not run cleanup logic on component unmount. This meant that the feed client would not be disposed when the component unmounted, potentially leading to memory leaks or lingering subscriptions. By switching to `useEffect`, we ensure that the feed client is properly initialized and disposed of both when dependencies change and when the component unmounts, resulting in more predictable and stable behavior.

### Tasks
[KNO-8435](https://linear.app/knock/issue/KNO-8435/[sdk]-investigate-potential-memory-leak-with-useref-in-feed-client)

### DIFF
--- a/.changeset/great-fans-post.md
+++ b/.changeset/great-fans-post.md
@@ -1,0 +1,5 @@
+---
+"@knocklabs/react-core": patch
+---
+
+fixes memory leak when unmounting the useNotifications hook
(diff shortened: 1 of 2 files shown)


kylemcd:
* **#495** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/495?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/495?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

cjbell (approved):

PR #504, opened 2025-04-28
fix: quality of life improvements

cjbell:
- revert the memory leak fix from #495 
- fix typescript issue with `buttonRef`
- fix typescript issue with api key and feed id
- add check that feedId is a UUID in the feed constructor

### DIFF
--- a/packages/client/src/knock.ts
+++ b/packages/client/src/knock.ts
@@ -130,8 +130,8 @@ class Knock {
     }
   }
 
-  log(message: string) {
-    if (this.logLevel === "debug") {
+  log(message: string, force = false) {
+    if (this.logLevel === "debug" || force) {
       console.log(`[Knock] ${message}`);
     }
   }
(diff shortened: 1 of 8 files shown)


packages/client/src/knock.ts
```diff
@@ -130,8 +130,8 @@ class Knock {
     }
   }
 
-  log(message: string) {
-    if (this.logLevel === "debug") {
+  log(message: string, force = false) {
```
connorlindsey, line 133: Any reason to use this vs. setting the log level?

packages/react-core/src/modules/feed/context/KnockFeedProvider.tsx
```diff
@@ -26,7 +26,7 @@ const KnockFeedContext = React.createContext<
 
 export interface KnockFeedProviderProps {
   // Feed props
-  feedId: string;
+  feedId: string | undefined;
```
connorlindsey, line 29: Uber nit: could use `?:` to be consistent with other props that can be undefined like color mode.

packages/client/src/knock.ts
cjbell: I guess i could have set log-level and then set it back – just felt like it was cleaner to have an override

PR #522, opened 2025-05-05
feat: add ability to override slack scopes

cjbell:
Changed `useSlackAuth` to:

- add new `options` param
- introduce ability to override `scopes`
- change api to break out `additionalScopes` under options

### DIFF
--- a/packages/react-core/src/modules/slack/hooks/useSlackAuth.ts
+++ b/packages/react-core/src/modules/slack/hooks/useSlackAuth.ts
@@ -24,21 +24,38 @@ type UseSlackAuthOptions = {
   additionalScopes?: string[];
 };
 
+// Here we normalize the options to be a single object with scopes and additionalScopes
+// The "options" parameter can be an array of scopes, an object with scopes and additionalScopes, or undefined
+// We handle the array case because it was the previous way to pass options so we're being backward compatible
+function normalizeOptions(options?: UseSlackAuthOptions | string[]): {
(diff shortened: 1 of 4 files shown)


JEverhart383 (approved):
LGTM

packages/react-core/src/modules/slack/hooks/useSlackAuth.ts
```diff
@@ -17,19 +17,28 @@ type UseSlackAuthOutput = {
...
+  scopes?: string[];
+  // Additional scopes to add to the default scopes
+  additionalScopes?: string[];
+};
+
 function useSlackAuth(
   slackClientId: string,
   redirectUrl?: string,
-  additionalScopes?: string[],
+  options?: UseSlackAuthOptions,
```
mattmikolay, line 30: Nit: changing this from `string[]` to an object is a breaking change. If we’re strictly following semantic versioning, this should not be a patch bump.

What do you think about making this `options?: string[] | UseSlackAuthOptions` and then maintaining the existing behavior when the type is `string[]` so that we keep backwards compatibility?

mattmikolay (approved):

packages/react-core/src/modules/slack/hooks/useSlackAuth.ts
cjbell: Made this backwards compatible as suggested 👍

PR #541, opened 2025-05-13
feat: support image field in modal and card components

thomaswhyyou:
### What changed?

We are adding an optional image field to the Card and Modal out-of-the-box (OOTB) message types ([here](https://github.com/knocklabs/control/pull/5057)), which means our OOTB components need to be able to render it accordingly so this PR adds an image field to the Card and Modal components.

Also related: https://github.com/knocklabs/control/pull/5073

### Tasks
[KNO-8613](https://linear.app/knock/issue/KNO-8613/control-add-an-image-field-to-card-and-modal-ootb-schemas)

### Screenshots or Loom

**Card:** 
CleanShot 2025-05-13 at 15 27 38@2x

**Modal:**
CleanShot 2025-05-13 at 16 26 54@2x

### DIFF
--- a/packages/react/src/modules/guide/components/Banner/Banner.tsx
+++ b/packages/react/src/modules/guide/components/Banner/Banner.tsx
@@ -3,11 +3,7 @@ import clsx from "clsx";
 import React from "react";
 
 import { maybeNavigateToUrlWithDelay } from "../helpers";
-import {
-  ButtonContent,
-  TargetButton,
-  TargetButtonWithGuideContext,
-} from "../types";
+import { ButtonContent, TargetButton, TargetButtonWithGuide } from "../types";
 
 import "./styles.css";
 
@@ -181,10 +177,7 @@ DefaultView.displayName = "BannerView.Default";
 
 type BannerProps = {
   guideKey?: string;
-  onButtonClick?: (
-    e: React.MouseEvent,
-    target: TargetButtonWithGuideContext,
-  ) => void;
+  onButtonClick?: (e: React.MouseEvent, target: TargetButtonWithGuide) => void;
 };
 
 export const Banner: React.FC<BannerProps> = ({ guideKey, onButtonClick }) => {
--- a/packages/react/src/modules/guide/components/Card/Card.tsx
+++ b/packages/react/src/modules/guide/components/Card/Card.tsx
@@ -92,8 +92,8 @@ const Img: React.FC<
   return (
     <img
       className={clsx("knock-guide-card__img", className)}
-      {...props}
       alt={alt || ""}
+      {...props}
     >
       {children}
     </img>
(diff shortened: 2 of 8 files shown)


thomaswhyyou:
* **#541** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/541?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/541?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/react/src/modules/guide/components/Banner/Banner.tsx
```diff
@@ -3,11 +3,7 @@ import clsx from "clsx";
 import React from "react";
 
 import { maybeNavigateToUrlWithDelay } from "../helpers";
-import {
-  ButtonContent,
-  TargetButton,
-  TargetButtonWithGuideContext,
-} from "../types";
+import { ButtonContent, TargetButton, TargetButtonWithGuide } from "../types";
```
thomaswhyyou, line 6: Unrelated, but I changed the type name to `TargetButtonWithGuide` from `TargetButtonWithGuideContext`

packages/react/src/modules/guide/components/Card/Card.tsx
```diff
@@ -170,47 +196,70 @@ const DefaultView: React.FC<{
...
-                const { text, action } = content.secondary_button!;
-                onButtonClick(e, { name: "secondary_button", text, action });
-              }
-            }}
-          />
-        )}
-      </Actions>
+        </a>
+      )}
+      {(content.primary_button || content.secondary_button) && (
```
thomaswhyyou, line 219: Nothing changed with buttons, but right now we are still rendering `<Actions />` even if there are no buttons and this adds extra spacing at the bottom due to the gap attribute in the container. We should only render it if there is at least one button.

packages/react/src/modules/guide/components/Modal/Modal.tsx
```diff
@@ -208,32 +229,59 @@ const DefaultView: React.FC<{
...
-                  onButtonClick(e, { name: "primary_button", text, action });
-                }
-              }}
-            />
-          )}
-        </Actions>
+          </a>
+        )}
+
+        {(content.primary_button || content.secondary_button) && (
```
thomaswhyyou, line 253: Same here.

packages/react/src/modules/guide/components/Modal/Modal.tsx
```diff
@@ -108,6 +111,21 @@ const Body: React.FC<{ body: string } & React.ComponentPropsWithRef<"div">> = ({
 };
 Body.displayName = "ModalView.Body";
 
+const Img: React.FC<
```
connorlindsey, line 114: Worth extracting any of this into shared image components that take a type? Seems fine duplicating them at this point.

connorlindsey (approved):

packages/react/src/modules/guide/components/Modal/Modal.tsx
thomaswhyyou: Yeah I think keeping them separate feels fine to me at this time, not a ton going on with this component to make me feel like we need to abstract things away.

PR #543, opened 2025-05-14
chore(KNO-8638): add tests for useNotifications hook

mattmikolay:
### Description

I’m working on refactoring the `useNotifications` hook to fix a memory leak that occurs due to the fact that the hook does not dispose of feed clients on unmount. I’m adding some test coverage to make sure I don’t break existing functionality.

### Checklist
- [X] Tests have been added for new features or major refactors to existing features.

PR #544, opened 2025-05-15
fix(KNO-8595): dispose feed on unmount in useNotifications hook

mattmikolay:
### Description

The current implementation of `useNotifications` has a memory leak: when the component using the `useNotifications` hook unmounts, the hook does not dispose of the feed it previously initialized.

This PR refactors `useNotifications` so that feed clients are properly initialized and disposed. I relied heavily on the unit tests I added in #543 to make sure I didn’t break existing behavior, and performed some manual testing with an example app in addition.

Please note my comment about the extra feed client created when strict mode is enabled. [Here’s a Loom with some additional thoughts](https://www.loom.com/share/b8e045b2870d4754856cbd0cc4656706?sid=407e62da-3f3e-4e75-bf6b-826b2af9d270).

### Checklist


mattmikolay:
* **#544** <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/544?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/knocklabs/javascript/544?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/react-core/src/modules/feed/hooks/useNotifications.ts
```diff
@@ -1,36 +1,69 @@
...
+    [knock],
+  );
+
   const stableOptions = useStableOptions(options);
+  const [state, setState] = useState<State>(() => ({
+    // FIXME In strict mode, useState initializer functions are called twice,
+    // which results in one extra instance of the feed client being created
+    // and not disposed of. This only affects strict mode.
+    feedClient: initFeedClient(feedChannelId, stableOptions),
+    options: stableOptions,
```
mattmikolay, line 37: From my testing, the extra instance of the feed client created in strict mode doesn’t actually cause any issues. Fixing this would require a non-trivial amount of refactoring to code in our client SDK.

In my opinion, the way this hook _should_ work is something like this:

```tsx
function useNotifications() {
    const [feed, setFeed] = useState(new Feed());

    useEffect(() => {
        feedClientManager.register(feed);
        return () => feedClientManager.unregister(feed);
    }, [feedClient]);

    return feedClient;
}
```

I’m glossing over a lot of details here, and I’m assuming the existence of a `feedClientManager`, but the point is, the created feed shouldn’t actually do _anything_ until it’s registered within the `useEffect` via `feedClientManager.register(feed)`.

kylemcd (approved):
Thanks for the super thorough explanation. This looks great to me and I'm glad we got some test cases in here for this functionality! :chefkiss:

packages/react-core/src/modules/feed/hooks/useNotifications.ts
cjbell: Yeah this is a really good point. The feedclient as it's designed right now is doing too much on init and agree that it should be stateless until it's needed. It was always really hard to fit into the hooks, hence the useRef dance, and that is a huge smell. I think we should incorporate this into a future client improvements brief.

cjbell (approved):

packages/react-core/src/modules/feed/hooks/useNotifications.ts
```diff
@@ -1,36 +1,69 @@
...
 
-    feedClientRef.current = knock.feeds.initialize(
-      feedChannelId,
-      stableOptions,
-    );
+    // Initialize a new feed client if the feed ID has changed,
+    // options have changed, or the current feed client has been disposed
+    const needsReinit =
+      state.feedClient.feedId !== feedChannelId ||
+      state.options !== stableOptions ||
```
cjbell, line 48: Should we do a shallow comparison here? Is this safe to just do like this?

cjbell (approved):

packages/react-core/src/modules/feed/hooks/useNotifications.ts
mattmikolay: Do you think this should be added to [the existing Javascript SDK visibility & stability brief](https://www.notion.so/knockapp/Javascript-SDK-visibility-stability-1f4c5a20997880469909d5427d45d7da?pvs=4) Kyle just shared? Or as a new brief entirely?

PR #626, opened 2025-07-08
feat: support inline identify

kylemcd:
### Description
** Describe what, why and how of the changes clearly and concisely. Add any additional useful context or info, as necessary. **

### Todos
** List any todo items necessary before merging, if any. Delete if none. **
- [ ] Sample todo item 1
- [ ] Sample todo item 2

### Checklist
- [ ] Tests have been added for new features or major refactors to existing features.

### Screenshots or videos
** Attach any screenshots or recordings to visually illustrate the changes, as necessary. Delete if not relevant. **


packages/client/src/knock.ts
```diff
@@ -103,6 +120,15 @@ class Knock {
       this.log("Reinitialized real-time connections");
     }
 
+    // Inline identify the user if we've been given an object with an id.
+    if (
```
connorlindsey, line 124: Did we discuss whether this should be opt-in or not? I'm good with it being the default but wanted to double check.

connorlindsey (approved):
Looks great!

packages/client/src/knock.ts
kylemcd: I assume we'd just want this to be the default. This wouldn't actually work until the user migrates to passing the user object thought.

What do you think about this being default vs opting in @cjbell

PR #640, opened 2025-07-10
chore: update example apps to use inline identify + fix app dir example build errors

kylemcd:
### Description
- Updates example apps to use inline identify pattern, i.e. `user={{ id: user.id }}`.
- Fixes build issues with `nextjs-app-dir-example`. I don't think this ever built correctly, but needed to run a build here to verify the updates from inline identify didn't break anything.

### DIFF
--- a/examples/guide-example/src/App.tsx
+++ b/examples/guide-example/src/App.tsx
@@ -71,7 +71,7 @@ function App() {
             <Route index element={<Page title="home" />} />
             <Route path="/produce" element={<Page title="Produce" />} />
             <Route path="/meat" element={<Page title="Meat" />} />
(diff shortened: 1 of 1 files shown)


examples/guide-example/src/App.tsx
```diff
@@ -72,9 +72,18 @@ function App() {
             <Route path="/produce" element={<Page title="Produce" />} />
             <Route path="/meat" element={<Page title="Meat" />} />
             <Route path="/seafood" element={<Page title="Seatfood" />} />
```
mattmikolay, line 74: LOL `"Seatfood"` 🤔

mattmikolay (approved):

PR #670, opened 2025-07-24
feat: pass sdk version as user-agent header

cjbell:
### Description

Adapts the User-Agent header sent by the SDK to pass a header that matches our stainless-generated SDKs so we can use them in analysis

Header looks like: `Knock/ClientJS 0.15.0`


.changeset/version-plugin.ts
```diff
@@ -0,0 +1,20 @@
+const fs = require("fs");
+const path = require("path");
+
+module.exports = {
```
kylemcd, line 4: i think you can just import the version from the package.json version key yeah?

cjbell: yeah and had that in past version, but wasn't sure if there's any negative in including that in the bundle then? not a huge json file but still a lot of extra data that's not really necessary

kylemcd: `package.json` is required to be in the bundle already

kylemcd (approved):
Looks good, don't forget to add a changeset 🚢

meryldakin:
KNO-8897: https://linear.app/knock/issue/KNO-8897/sdks-add-an-sdk-version-number-request-header

PR #680, opened 2025-07-31
feat: add useGuides hook

thomaswhyyou:
### Description
Add the `useGuides` hook, which will return any matching selections without applying the group limit of one-at-a-time display or throttle settings.

### DIFF
--- a/packages/client/src/clients/guide/client.ts
+++ b/packages/client/src/clients/guide/client.ts
@@ -358,11 +358,30 @@ export class KnockGuideClient {
   // Store selector
   //
 
+  selectGuides(state: StoreState, filters: SelectFilterParams = {}) {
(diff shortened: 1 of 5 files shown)


packages/client/src/clients/guide/client.ts
```diff
@@ -359,11 +359,32 @@ export class KnockGuideClient {
...
+    this.knock.log(`[Guide] Selecting guides for: ${formatFilters(filters)}`);
+
+    const result = select(state, filters);
+
+    if (result.size === 0) {
+      this.knock.log("[Guide] Selection returned zero result");
+      return [];
+    }
+
+    // console.log([...result.values()])
```
connorlindsey, line 375: Nit: Left over log & unfinished comment

connorlindsey (approved):

PR #721, opened 2025-08-19
feat: add Guide dev tools component

connorlindsey:
### Description
Adds a minimal Guides DevTool component for displaying debug info. Right now, this just shows the guide key that's selected and lets the user exit debug mode which clears the query param. 

### Checklist

### Screenshots or videos
CleanShot 2025-08-19 at 13 51 45@2x


packages/react/src/modules/guide/providers/KnockGuideProvider.tsx
```diff
@@ -0,0 +1,20 @@
+import {
+  KnockGuideProvider as KnockGuideProviderCore,
+  type KnockGuideProviderProps,
+} from "@knocklabs/react-core";
+import React from "react";
+
+import { GuideDevTools } from "../components";
+
+// Re-export the core KnockGuideProvider, so we can add React specific functionality
```
connorlindsey, line 9: I decided to add the devtools component here and re-export the KnockGuideProvider so that we can use Telegraph without impacting non-web React usage. This means we need to export some additional things like `KnockGuideProviderProps` from react-core.

packages/react/src/modules/ms-teams/theme.css
```diff
@@ -1,4 +1,4 @@
-@import "@telegraph/tokens/light";
+@import "@telegraph/tokens/default";
```
connorlindsey, line 1: MS Teams uses light mode, but the guide devtools uses the dark styles, so we need to import the default tokens and use `data-tgph-appearance` to swap (MS Teams won't be affected bc the default is light).

packages/react/src/modules/guide/components/GuideToolbar/GuideToolbar.tsx
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  const handleExit = () => {
+    const window = checkForWindow();
+    if (!window) {
+      return;
+    }
+
+    const url = new URL(window.location.href);
+    url.searchParams.delete("knock_guide_key");
```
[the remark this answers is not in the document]
connorlindsey: Yeah, I'll export and use the same ones

connorlindsey:
### Merge activity

* **Sep 3, 6:14 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/knocklabs/javascript/721).

PR #722, opened 2025-08-20
feat: add content type generic for guide step content

thomaswhyyou:
### Description

Adds a generic type parameter for typing guide step contents, which users can pass in when using the `useGuide(s)` hooks or calling the `selecteGuide(s)` methods in the guide client.

Note, there is a corresponding CLI PR (https://github.com/knocklabs/knock-cli/pull/575) to add a new `knock guide generate-types` command which generates types for all guides in a given environment automatically.

### DIFF
--- a/packages/client/src/clients/guide/types.ts
+++ b/packages/client/src/clients/guide/types.ts
@@ -4,6 +4,9 @@ import { GenericData } from "@knocklabs/types";
 // Fetch guides API
 //
 
+// eslint-disable-next-line
+export type Any = any;
+
 export interface StepMessageState {
   id: string;
   seen_at: string | null;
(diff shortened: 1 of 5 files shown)


packages/client/src/clients/guide/types.ts
```diff
@@ -4,6 +4,9 @@ import { GenericData } from "@knocklabs/types";
 // Fetch guides API
 //
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type Any = any;
```
cjbell, line 8: Wondering if `unknown` is better than any here

connorlindsey (approved):

PR #730, opened 2025-08-27
feat: handle guides live preview content

connorlindsey:
### Description
Handles live preview update events and uses preview content if available for the given guide.

### Checklist

### DIFF
--- a/packages/client/src/clients/guide/client.ts
+++ b/packages/client/src/clients/guide/client.ts
@@ -28,6 +28,7 @@ import {
   GuideData,
   GuideGroupAddedEvent,
   GuideGroupUpdatedEvent,
+  GuideLivePreviewUpdatedEvent,
   GuideRemovedEvent,
   GuideSocketEvent,
   GuideStepData,
@@ -58,7 +59,10 @@ const DEFAULT_COUNTER_INCREMENT_INTERVAL = 30 * 1000; // in milliseconds
 const SUBSCRIBE_RETRY_LIMIT = 3;
 
 // Debug query param keys
-const DEBUG_GUIDE_KEY_PARAM = "knock_guide_key";
+export const DEBUG_QUERY_PARAMS = {
+  GUIDE_KEY: "knock_guide_key",
+  PREVIEW_SESSION_ID: "knock_preview_session_id",
+};
 
 // Return the global window object if defined, so to safely guard against SSR.
 const checkForWindow = () => {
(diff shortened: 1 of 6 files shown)


packages/client/src/clients/guide/client.ts
```diff
@@ -989,15 +1019,11 @@ export class KnockGuideClient {
...
-    // If entering/exiting debug mode, refetch guides and resubscribe to the
+    // If debug state has changed, refetch guides and resubscribe to the
     // websocket channel.
-    if (
-      Boolean(currentDebugParams.forcedGuideKey) !==
-      Boolean(newDebugParams.forcedGuideKey)
-    ) {
-      this.knock.log(
-        `[Guide] ${newDebugParams.forcedGuideKey ? "Entering" : "Exiting"} debug mode`,
-      );
```
thomaswhyyou, line 1000: Not a deal breaker but may be good to log something.

packages/client/src/clients/guide/client.ts
```diff
@@ -989,15 +1019,11 @@ export class KnockGuideClient {
     const newDebugParams = detectDebugParams();
     this.setLocation(href, { debug: newDebugParams });
```
thomaswhyyou, line 1020: In setLocation(), maybe we should also clear out `previewGuides` if `preivew_session_id` is being unset?

packages/client/src/clients/guide/client.ts
```diff
@@ -99,9 +106,17 @@ const select = (state: StoreState, filters: SelectFilterParams = {}) => {
...
-    const guide = state.guides[guideKey];
+    let guide = state.guides[guideKey];
     if (!guide) continue;
 
+    // Use preview guide if it exists and matches the forced guide key
+    if (
+      state.debug.forcedGuideKey === guideKey &&
+      state.previewGuides[guideKey]
+    ) {
+      guide = state.previewGuides[guideKey];
```
thomaswhyyou, line 117: It probably doesn't make much difference in reality, but we are swapping out the guide ahead of `predicate()` which means we are filtering in predicate() with the preview guide vs the regular one. Could we move this down below predicate() but above result.set(), so we continue to use the real guide and only swap it out at the end?

packages/client/src/clients/guide/client.ts
```diff
@@ -989,15 +1019,11 @@ export class KnockGuideClient {
...
     // websocket channel.
-    if (
-      Boolean(currentDebugParams.forcedGuideKey) !==
-      Boolean(newDebugParams.forcedGuideKey)
-    ) {
-      this.knock.log(
-        `[Guide] ${newDebugParams.forcedGuideKey ? "Entering" : "Exiting"} debug mode`,
-      );
+    const debugStateChanged =
+      JSON.stringify(currentDebugParams) !== JSON.stringify(newDebugParams);
```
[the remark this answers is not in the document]
thomaswhyyou: > since we actually only care whether or not guide key is present or not (since either way, we're just sending 'force all guides' to SB)

Double checking, don't you send `preview_session_id` for subscribe() also in Line 366 though?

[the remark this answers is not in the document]
connorlindsey: I meant that specifically when comparing the guide key we just care about presence, but yeah, we still want to compare the rest of the debug state, here's what I added:

```
  private checkDebugStateChanged(a: DebugState, b: DebugState): boolean {
    return (
      Boolean(a.forcedGuideKey) !== Boolean(b.forcedGuideKey) ||
      a.previewSessionId !== b.previewSessionId
    );
  }
```

PR #743, opened 2025-09-09
fix: filter out inactive guides in preview mode

connorlindsey:
### Description
Filters out inactive guides in preview mode

### Checklist

### DIFF
--- a/packages/client/src/clients/guide/client.ts
+++ b/packages/client/src/clients/guide/client.ts
@@ -160,6 +160,13 @@ const predicate = (
     return true;
   }
 
+  // If we're in debug mode, replicate some server side filtering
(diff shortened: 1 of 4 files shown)


packages/client/src/clients/guide/client.ts
```diff
@@ -160,6 +160,13 @@ const predicate = (
     return true;
   }
 
+  // If we're in debug mode, replicate some server side filtering
+  if (debug.forcedGuideKey) {
+    if (!guide.active) {
+      return false;
+    }
+  }
```
thomaswhyyou, line 168:
```suggestion
  if (!guide.active) {
    return false;
  }
```
I think you can just apply this logic generally regardless of normal vs debug mode.

PR #745, opened 2025-09-12
feat: add auto socket manager to guide provider and client

connorlindsey:
### Description
- Abstracts auto socket manager into its own class
- Uses in feed and guide client

I tested this locally while live previewing a guide and it successfully disconnects and reconnects and still receives live preview updates on the re-established connection. 

### Checklist

### Screenshots or videos
CleanShot 2025-09-12 at 10 09 51

### DIFF
--- a/.changeset/polite-bees-serve.md
+++ b/.changeset/polite-bees-serve.md
@@ -1,0 +1,6 @@
+---
+"@knocklabs/react-core": patch
+"@knocklabs/client": patch
+---
+
+feat: add auto socket manager to guide provider and client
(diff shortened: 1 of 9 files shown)


thomaswhyyou (commented):
Sorry for the delay on reviewing this. I just spent some time looking through this PR, and there are a couple aspects of the current implementation (which I noted) that is giving me a pause.. I think the fact that we are sharing a single socket connection between Feed and KnockGuideClient makes it quite tricky to handle document visibility independently like this.

Happy to discuss further, but I think this requires some additional thinking and more substantial refactoring which makes it a bit unclear whether we could have it ready for GA.

thomaswhyyou:
Closing this for now, we will revisit this in the future.

PR #751, opened 2025-09-15
feat(feed): optimistically update when unarchiving an item

joshuaellis:
### Description
** Describe what, why and how of the changes clearly and concisely. Add any additional useful context or info, as necessary. **

If you had the feed default option to `archived: 'only'` then unarchiving an item is not optimistically removed from the list and would require the user to call `feed.fetch` again.  Instead we now use the same optimistic removing logic (should we consider moving this to it's own private fn?) as `markAsArchived` where if the archived option is set to `only` we remove the item.

I've written two new tests, the `markAsArchived` will pass on `main` but the `markAsUnarchived` will fail. Now they both pass.

### Screenshots or videos
** Attach any screenshots or recordings to visually illustrate the changes, as necessary. Delete if not relevant. **

This is a video of the bug in question:

https://github.com/user-attachments/assets/1482f9bd-cf3e-48c9-a97f-5cfa812a48c9

I can see that the item is recognised as `unarchived` because the button changes to archived as expected.

### DIFF
--- a/packages/client/src/clients/feed/feed.ts
+++ b/packages/client/src/clients/feed/feed.ts
@@ -472,9 +472,46 @@ class Feed {
   }
 
   async markAsUnarchived(itemOrItems: FeedItemOrItems) {
-    this.optimisticallyPerformStatusUpdate(itemOrItems, "unarchived", {
-      archived_at: null,
-    });
+    const state = this.store.getState();
+
+    const items = Array.isArray(itemOrItems) ? itemOrItems : [itemOrItems];
+
+    const itemIds: string[] = items.map((item) => item.id);
+
+    const shouldOptimisticallyRemoveItems =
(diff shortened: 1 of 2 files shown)


connorlindsey (approved):
Hey, this is excellent! Thanks for the contribution and matching the existing pattern we have with `markAsArchived`. I just pushed some minor changes and we'll get this merged and into the next release!

joshuaellis:
> Hey, this is excellent! Thanks for the contribution and matching the existing pattern we have with `markAsArchived`. I just pushed some minor changes and we'll get this merged and into the next release!

Thanks for doing those last bits to get it over the line 💪🏼

connorlindsey:
@joshuaellis Of course! These changes are released in v0.8.5 (https://www.npmjs.com/package/@knocklabs/react/v/0.8.5)

PR #758, opened 2025-09-22
fix: persist guide debug state in local storage

connorlindsey:
### Description
Persists debug state to local storage so that it persists across navigation changes (e.g. if you land on an authenticated page and get redirected, it'll persist the debug state). 

### Checklist

### Screenshots or videos
https://www.loom.com/share/3ceeb82f0c8c41fe9fca931e59517303?sid=f401b28b-ba93-4317-95b4-9248ab7de3c8

---

> [!NOTE]
> Persists guide debug params to localStorage across navigations, adds an API to exit debug mode (clearing storage/state/URL), updates the toolbar to use it, and adds tests.
> 
> - **Guides (client)**:
>   - Persist debug params (`knock_guide_key`, `knock_preview_session_id`) from URL to `localStorage`; read them when URL params are absent; add safe JSON parsing.
>   - New `exitDebugMode()` clears `localStorage`, store debug state/preview guides, and removes debug query params from the URL.
> - **React UI**:
>   - `GuideToolbar` now calls `client.exitDebugMode()` instead of manually editing URL params.
> - **Tests**:
>   - Add coverage for `localStorage` error handling and persistence during initialization and `handleLocationChange`.
> - **Tooling**:
>   - Bump `vitest` to `3.2.4` across packages.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5e3b0d717f94e051acca5d9bc1fc142906aa2433. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


packages/client/src/clients/guide/client.ts
```diff
@@ -76,18 +82,60 @@ const checkForWindow = () => {
...
+        win.localStorage.setItem(
+          DEBUG_STORAGE_KEYS.GUIDE_KEY,
+          urlGuideKey || "",
+        );
+        win.localStorage.setItem(
+          DEBUG_STORAGE_KEYS.PREVIEW_SESSION_ID,
+          urlPreviewSessionId || "",
+        );
+      } catch {
+        // Silently fail in privacy mode
```
connorlindsey, line 111: Seems rare, but local storage can throw in some browsers or if storage is full. 

Also, I opted to just repeat these trycatch & window checks since we only have a few local storage operations, but we could refactor these into a shared util to clean it up.

packages/client/src/clients/guide/client.ts
```diff
@@ -470,6 +518,42 @@ export class KnockGuideClient {
     });
   }
 
+  exitDebugMode() {
```
connorlindsey, line 521: Adding this method here so that the GuideToolbar doesn't have to be concerned with clearing the url/local storage/etc.

package.json
```diff
@@ -49,7 +49,7 @@
     "@vitest/coverage-v8": "^3.2.4",
     "prettier": "^3.5.3",
     "turbo": "^2.5.4",
-    "vitest": "^3.1.1"
+    "vitest": "^3.2.4"
```
connorlindsey, line 52: Tests weren't running because Vitest and vitest/coverage versions weren't matching

packages/client/src/clients/guide/client.ts
```diff
@@ -76,18 +78,59 @@ const checkForWindow = () => {
...
-  return { forcedGuideKey, previewSessionId };
+  // Check local storage if no URL params
+  let storedGuideKey = null;
+  let storedPreviewSessionId = null;
+
+  if (win.localStorage) {
+    try {
+      const storedDebugState = win.localStorage.getItem(DEBUG_STORAGE_KEY);
+      if (storedDebugState) {
+        const parsedDebugState = JSON.parse(storedDebugState);
```
thomaswhyyou, line 121: Could we create a helper like `safeJsonParse` that wraps this in try/catch, and return `undefined` in case of any parse error instead of throwing so we never crash from this?

thomaswhyyou (commented):
One follow-up thing.

packages/client/src/clients/guide/client.ts
connorlindsey: Yep, good suggestion, I added a method to make sure we always return `DebugState` after parsing

PR #759, opened 2025-09-24
fix(KNO-9854): handle promise rejection when identifying inline

mattmikolay:
### Description

`this.users.identify(properties)` returns a promise. This PR adds a `catch` so that we handle any rejection.

I looked into enabling [the `no-floating-promises` ESLint rule](https://typescript-eslint.io/rules/no-floating-promises/), but that requires some updates to our ESLint and TypeScript config files, so I’d like to ship this separately.

### Checklist
- [X] Tests have been added for new features or major refactors to existing features.

### DIFF
--- a/.changeset/short-masks-decide.md
+++ b/.changeset/short-masks-decide.md
@@ -1,0 +1,5 @@
+---
+"@knocklabs/client": patch
+---
+
+fix: handle promise rejection when identifying a user inline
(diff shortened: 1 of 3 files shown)


kylemcd (approved):
approving to unblock the customer but: does this need to be implemented in more places / could it just be handled inside of users.identify instead of each call site?

PR #795, opened 2025-10-30
feat: add next specific helper components for guides to detect location changes

thomaswhyyou:
### Description
This is a re-attempt of #789, to introduce Next js specific helper components for guides to detect route changes. 

The previous attempt was causing a [module resolution error](https://knocklabs.slack.com/archives/C079G6X3DD4/p1761759401293979) and breaking the build in our dashboard by simply upgrading `@knocklabs/react` to the latest version with the nextjs helpers, so we rolled that back.

This PR contains the same helper modules as #789, but with a few changes to how we organize/build/import them:
1. Emit multi-entry builds: the default build for react via the root entry ("@knocklabs/react"), and an optional one for next specific modules like ("@knocklabs/react/next"). See more details [here](https://github.com/knocklabs/javascript/pull/795/files#r2478547766) and [here](https://github.com/knocklabs/javascript/pull/795/files#r2478560460).
2. Mark `next/*` packages as "external" in vite's rollup config, so we don't bundle them and defer to consumer's next app to resolve at runtime. See more details [here](https://github.com/knocklabs/javascript/pull/795/files#r2478526074).
3. Specify `.js` suffix in next/* import paths. See more details [here](https://github.com/knocklabs/javascript/pull/795/files#r2478716544).

---

> [!NOTE]
> Introduces Next.js-specific guide location sensors under `@knocklabs/react/next`, updates build/export config, and renames client cleanup method for location listeners.
> 
> - **React**
>   - **Next-specific helpers**: Add `KnockGuideLocationSensor` with `PagesRouter` and `AppRouter` in `packages/react/src/next/*` for route-based guide location tracking.
>   - **Exports/build**: Add secondary entry point `@knocklabs/react/next`; switch to multi-entry Vite build (`index`, `next`); mark `next` and `next/*` as externals; keep CSS handling; add optional `next` peer dependency and devDependency.
> - **Client**
>   - **Location listeners**: Rename `removeEventListeners` to `removeLocationChangeEventListeners` and use in `cleanup`; add log in `.handleLocationChange`.
> - **Release**
>   - Changeset: patch bumps for `@knocklabs/client` and `@knocklabs/react`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 3e1c329502d06c6bb58af84ae51391a7a9b3810d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #801, opened 2025-11-05
feat: add guide location sensor helper for tanstack router

thomaswhyyou:
### Description
Add `<KnockGuideLocationSensor />` helper component for tanstack router, similar to what we have done for Next js [here](https://github.com/knocklabs/javascript/pull/795).

---

> [!NOTE]
> Adds `KnockGuideLocationSensor` for Tanstack Router and exposes it via a new `@knocklabs/react/tanstack` export, with build and dependency updates.
> 
> - **@knocklabs/react**:
>   - **Tanstack Router integration**:
>     - Add `KnockGuideLocationSensor` in `src/tanstack/KnockGuideLocationSensor.tsx` to sync guide client location from router `href` and clear prior listeners.
>     - Export via new entry `@knocklabs/react/tanstack` (`src/tanstack/index.ts`).
>   - **Build/Exports**:
>     - Add `./tanstack` to `package.json` exports and Vite lib entries; mark `@tanstack/react-router` as external in Rollup.
>   - **Dependencies**:
>     - Add `@tanstack/react-router` as optional peer and dev dependency.
> - **Release**:
>   - Add changeset for a patch bump.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 7250bb6cbdd4da78f68a23853710ec888480d12e. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/.changeset/plain-onions-relate.md
+++ b/.changeset/plain-onions-relate.md
@@ -1,0 +1,5 @@
+---
+"@knocklabs/react": patch
+---
+
+[guide] add KnockGuideLocationSensor helper component for Tanstack Router


thomaswhyyou:
> Code looks good although I wonder if we really want to ship a bunch of these components, or if it'd be better to just have examples in the docs and expect users to implement this when they add the guide provider to their app.
> 
> Maintenance is kind of a pain then if we ever need to update the snippet. If we ship the component, users can just upgrade to get the latest version vs. having to check the docs and update their code.
> 
> I guess this is probably the right call, but wanted to bring it up just in case

Yeah I don't think we will want to ship framework specific helpers like this for every framework that we encounter with our customers, but rather just for a few that are dominant market leaders like Next js, Tanstack etc. And then for all other ones i.e. long tail, we can show example snippets for customers to implement themselves, like you were saying. I do think it makes sense to ship something which _most_ people can just drop in and go from the devex point of view.

PR #808, opened 2025-12-07
Add possibility to provide a locale when registering push token, make sure to not miss the last notification interaction and get expo push token if has permissions on boot

sync:
### Description
We have started relying on knock and are faced with some "issues" with the current sdk, wondering if you would be interested in applying some of those fixes?

### Checklist
- [ ] Tests have been added for new features or major refactors to existing features.

### DIFF
--- a/packages/react-native/src/modules/push/KnockPushNotificationProvider.tsx
+++ b/packages/react-native/src/modules/push/KnockPushNotificationProvider.tsx
@@ -12,6 +12,7 @@ export interface KnockPushNotificationContextType {
   registerPushTokenToChannel(
     token: string,
     channelId: string,
+    locale?: string,
   ): Promise<ChannelData | void>;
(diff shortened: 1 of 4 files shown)


packages/react-native/src/modules/push/KnockPushNotificationProvider.tsx
```diff
@@ -12,6 +12,7 @@ export interface KnockPushNotificationContextType {
   registerPushTokenToChannel(
     token: string,
     channelId: string,
+    locale?: string,
```
sync: the locale in our app can be overridden (not system wide) and at the moment there is no way to tell knock about the custom user's locale

andy-knock: Looks reasonable.

PR #809, opened 2025-12-08
fix(kno-10550): channel cache in slack and teams

emisilvacab:
### Description

Fixed channel cache issue in the SlackKit and also applied the same fix for the ms teams. 

The problem was that `useSlackChannels` hook was using SWR for fetching data with a cache key of `["SLACK_CHANNELS", cursor]`, this key didn’t include the `tenantId` or `knockSlackChannelId`. So when the user switched workspaces (different tenant) or disconnected from one slack server and connected to a new one, SWR would return the cached data from the previous connection since the cache key was the same.

### Checklist

### Screenshots or videos
Vibecoded a small example to show this by switching tenants

#### Before
https://github.com/user-attachments/assets/2503a8bf-ebf8-45a6-bfca-3a691a41fe27

#### After
https://github.com/user-attachments/assets/0adab4d2-eee8-43eb-890b-b2908fd00536

### DIFF
--- a/packages/react-core/test/ms-teams/useMsTeamsTeams.test.ts
+++ b/packages/react-core/test/ms-teams/useMsTeamsTeams.test.ts
@@ -125,12 +125,12 @@ describe("useMsTeamsTeams", () => {
       {
         ms_teams_teams: [
           { id: "1", displayName: "Team Alpha" },
-          null as any,
+          null,
           { id: "2", displayName: "Team Beta" },
-          undefined as any,
-          false as any,
-          0 as any,
-        ],
+          undefined,
+          false,
+          0,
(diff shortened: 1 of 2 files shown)


emisilvacab:
We're using `useSWRInfinite` for paginated data but i've noticed that the manual cache-clearing `useEffect` with `useRef` tracking is redundant so I'll remove it. By including `tenantId` and `knockSlackChannelId` in the cache key, swr automatically treats different tenants as separate cache entries. When these values change, the key changes, and swr handles cache isolation automatically.

kylemcd (commented):
Can we cover the [missing lines](https://github.com/knocklabs/javascript/pull/809#issuecomment-3638521625) and this should be good to go!

PR #821, opened 2026-01-13
fix: KnockGuideStep type definition to return a promise from async methods

thomaswhyyou:
### Description

Updates the `KnockGuideStep` type as reported here: https://knocklabs.slack.com/archives/C079G6X3DD4/p1768330901136269?thread_ts=1767880973.264099&cid=C079G6X3DD4

The mark* methods are async so should wrap its return value in a promise.

### DIFF
--- a/.changeset/yellow-worlds-smile.md
+++ b/.changeset/yellow-worlds-smile.md
@@ -1,0 +1,5 @@
+---
+"@knocklabs/client": patch
+---
+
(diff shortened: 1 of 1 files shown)


thomaswhyyou:
> Looks good. Can't remember release process off the dome, make sure this doesn't require a version number bump / changeset / something else before merging 👍

@MikeCarbone Yep just need a changeset to trigger the release pipeline.

PR #823, opened 2026-01-14
fix: rework expo push provider

cjbell:
### Description

Cleaned up the KnockExpoPushNotificationProvider to be more readable and try and clean up some of the useEffects that existed that I suspect were the source of the bugs we've been seeing.

Introduces more ways to access the expo projectId to ensure it's loaded correctly for all Expo projects from 46+.

### DIFF
--- a/packages/expo/src/modules/push/utils.ts
+++ b/packages/expo/src/modules/push/utils.ts
@@ -53,8 +53,7 @@ export function getProjectId(): string | null {
  * @returns The permission status string
  */
 export async function requestPushPermission(): Promise<string> {
-  const { status: existingStatus } =
-    await Notifications.getPermissionsAsync();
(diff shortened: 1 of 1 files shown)


packages/expo/src/modules/push/utils.ts
```diff
@@ -0,0 +1,165 @@
...
+  });
+}
+
+/**
+ * Check if the current environment supports push notifications.
+ * @returns true if push notifications are supported
+ */
+export function isPushNotificationSupported(): boolean {
+  return Device.isDevice;
+}
```
mattmikolay: I’m pretty sure the Android emulator can receive push notifications now. I haven’t tested recently with Expo, but I’ve successfully sent push notifications directly from FCM to the emulator.

mattmikolay (approved):
🚀

PR #825, opened 2026-01-15
fix: check win properties in browser code paths for react native

thomaswhyyou:
### Description

This PR aims to address the issue raised [here](https://knocklabs.slack.com/archives/C079G6X3DD4/p1768503837346129), where user is running into type errors in places accessing window properties in a react naive project. 

We already check for the presence of the `window` object in code paths meant for browser environment. However it seems that in React Native environments it's possible for window may be polyfilled and present but its properties are undefined (e.g. window.location). So in this PR we extend some of the if checks to cover the window properties we are accessing, in addition to the presence of window itself.

### DIFF
--- a/.changeset/tall-berries-smell.md
+++ b/.changeset/tall-berries-smell.md
@@ -1,0 +1,5 @@
+---
+"@knocklabs/client": patch
+---
+
+[guides] check window properties in browser specific code paths for react native


thomaswhyyou:
* **#825** <a href="https://app.graphite.com/github/pr/knocklabs/javascript/825?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/knocklabs/javascript/825?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #869, opened 2026-02-19
chore(PLA-2224): re-fetch guides when the existing socket is disconnected and reconnected

jareddellitt:
### Description
This will force re-fetch guides when the socket is reopened after a disconnection. This handles both deployments of the underlying sockets system and page visibility induced disconnections.

### Checklist

### DIFF
--- a/packages/client/src/clients/guide/client.ts
+++ b/packages/client/src/clients/guide/client.ts
@@ -465,8 +465,10 @@ export class KnockGuideClient {
 
     // Refetch guides on socket reconnect to pick up any changes that
     // occurred while disconnected (e.g. tab was hidden for a long time).
-    // Skip the first onOpen since that's the initial connection, not a reconnect.
-    let isFirstOpen = true;
+    // If the socket is already connected, there's no initial open event coming,
+    // so we don't need to skip anything. If it's not yet connected, skip the
+    // first onOpen since that's the initial connection, not a reconnect.
+    let isFirstOpen = !this.socket.isConnected();
     this.socketOpenListenerId = this.socket.onOpen(() => {
       if (isFirstOpen) {
         isFirstOpen = false;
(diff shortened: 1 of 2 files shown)


packages/client/src/clients/guide/client.ts
```diff
@@ -416,6 +417,12 @@ export class KnockGuideClient {
       this.unsubscribe();
     }
 
+    // If there's an existing open listener, then remove it.
+    if (this.socketOpenListenerId) {
+      this.socket.off([this.socketOpenListenerId]);
+      this.socketOpenListenerId = undefined;
+    }
+
```
thomaswhyyou: Could we move this to a private method and re-use here and in unsubscribe?

packages/client/src/clients/guide/client.ts
```diff
@@ -455,6 +462,21 @@ export class KnockGuideClient {
...
     // Track the joined channel.
     this.socketChannel = newChannel;
+
+    // Refetch guides on socket reconnect to pick up any changes that
+    // occurred while disconnected (e.g. tab was hidden for a long time).
+    // If the socket is already connected, there's no initial open event coming,
+    // so we don't need to skip anything. If it's not yet connected, skip the
+    // first onOpen since that's the initial connection, not a reconnect.
+    let isFirstOpen = !this.socket.isConnected();
+    this.socketOpenListenerId = this.socket.onOpen(() => {
```
thomaswhyyou: Is it fair to assume this will fire when we deploy switchboard? I guess switchboard-sockets specifically?

If so, I kinda want to avoid triggering ALL connected clients to re-fetch at the same time from a deploy. Curious if you have any thoughts on targeting re-connect events from the tab visibility state change more granularly? Those will happen more organically and spread out across connected clients, which I think is would be the right thing to do. But i'm a bit hesitant to just blanket re-fetch based on every socket onOpen events.

PR #879, opened 2026-02-25
feat: add focusedGuideKeys debug setting to pin a guide

thomaswhyyou:
### Description

* Adds a `focusedGuideKeys` debug setting to the guide client, an object to hold guide keys to "focus" while in debug mode
  * Note, this effectively replaces the existing `forcedGuideKey` debug setting - with one difference being that `focusedGuideKeys` can hold more than one guide keys. For now we still limit to one guide key in the toolbar UI layer, but I wanted to make this change now in case we need to come back to extend this for useGuides (i.e. force rendering multiple guides with useGuides)
  * The plan is to clean up and remove `forcedGuideKey` after toolbar v2 ships.
* Adds a couple placeholder buttons so Krisna can take them into account for design refresh. Will come back in separate PRs to make them functional. 

### Screenshots or videos

https://github.com/user-attachments/assets/ed061a94-f1f7-4240-98e1-cea81a6c2ecc


thomaswhyyou:
@cursor review

PR #886, opened 2026-03-03
feat: read and write run config with local storage for guide toolbar v2

thomaswhyyou:
### Description

Read and write the guide toolbar run config via local storage, in addition to detecting them in the current url. 

This mirrors the behavior we currently have with "live preview" in the guide client, but for the guide toolbar.

### DIFF
--- a/packages/client/test/clients/guide/guide.test.ts
+++ b/packages/client/test/clients/guide/guide.test.ts
@@ -4486,6 +4486,7 @@ describe("KnockGuideClient", () => {
       vi.spyOn(client, "subscribe").mockImplementation(() => {});
 
       const clearGroupStageSpy = vi.spyOn(
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
         client as any,
         "clearGroupStage",
       );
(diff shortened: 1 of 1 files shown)


packages/client/src/clients/guide/client.ts
```diff
@@ -596,6 +596,10 @@ export class KnockGuideClient {
     this.knock.log("[Guide] .setDebug()");
     const shouldRefetch = !this.store.state.debug?.debugging;
 
+    // Clear the stage before updating the store state to let the next render
+    // cycle run with a fresh group stage.
+    this.clearGroupStage();
```
thomaswhyyou: Unrelated, but right now pinning a modal guide (i.e. `forcedGuideKeys` debug setting) can get the toolbar into a weird state with the group stage (because subsequent clicks in the toolbar registers as a "click outside" for the modal, which fires a markAsArchived call, which then triggers the guide state store to updated while `forcedGuideKeys` is set...). 

So this just clears the group stage and allows each render cycle to resolve with a new one cleanly with each setDebug/unsetDebug call.

PR #1025, opened 2026-06-30
fix: handle missing features in feed settings response

### LINKED ISSUE #1018: Users with bad internet: can't access property "branding_required", f.features is undefined
Image

    "@knocklabs/client": "^0.21.13",
    "@knocklabs/react": "^0.11.22",

abdtriedcoding:
### Description

Fixes #1018

Users on flaky networks can receive a successful (200) response from the feed settings endpoint with a partial or empty body (e.g. missing the `features` object). The SDK then crashed with:

`TypeError: can't access property "branding_required", v.features is undefined`

This PR:

- Normalizes feed settings in `useFeedSettings` so `features.branding_required` always defaults safely when missing
- Adds optional chaining on `features` in both web and React Native `NotificationFeed` components
- Adds a unit test for the partial-settings response case

### Todos

None.

### Checklist

### Test plan

- [x] `yarn test` — `useFeedSettings` tests pass (including new partial response case)
- [x] `yarn workspace @knocklabs/react-core type:check` passes
- [x] `yarn workspace @knocklabs/react-core lint` passes
- [x] `yarn workspace @knocklabs/react lint` passes
- [x] `yarn workspace @knocklabs/react-native lint` passes

### DIFF
--- a/.changeset/fix-feed-settings-features.md
+++ b/.changeset/fix-feed-settings-features.md
@@ -1,0 +1,7 @@
+---
+"@knocklabs/react-core": patch
+"@knocklabs/react": patch
+"@knocklabs/react-native": patch
+---
+
+Handle missing `features` in feed settings responses to prevent crashes on partial API responses.
--- a/packages/react-core/src/modules/feed/hooks/useFeedSettings.ts
+++ b/packages/react-core/src/modules/feed/hooks/useFeedSettings.ts
@@ -11,7 +11,7 @@ function useFeedSettings(feedClient: Feed): {
   settings: FeedSettings | null;
   loading: boolean;
 } {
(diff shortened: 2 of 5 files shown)


cjbell (approved):

kylemcd:
thanks @abdtriedcoding! pushed a couple edits on top:

- rebased onto main — `@knocklabs/client` just swapped axios for fetch (#1010), so this needed to sit on the new transport
- instead of defaulting `branding_required: false` when the body is missing `features`, we now leave settings null. a flaky 200 (captive portal/proxy) shouldn't silently hide required branding. guard is `response.statusCode === "ok" && response.body?.features` now, and updated the test to match

same crash fix, just doesn't assume a branding value we never got. thanks again!

PR #1030, opened 2026-07-08
fix(KNO-13857): reduce websocket reconnect load when connections can't recover

kylemcd:
### Description

Fixes [KNO-13857](https://linear.app/knock/issue/KNO-13857/javascript-reduce-ws-retry-mechanism). Follow-up to **INC-408**, where a customer rotated their public API key while deployed feed clients kept the old one. Each client's socket kept reconnecting with stale credentials, got a 403 on every upgrade, and retried forever — fleet-wide this grew websocket inbound volume ~1000% and saturated the sockets pods.

Browsers don't expose the HTTP status of a failed WS upgrade to JS — Phoenix only sees an abnormal close (1006) and reschedules — so a permanently-failing connection can never "give up" on its own. This PR makes a failing socket get *quieter over time* and makes background tabs stop entirely, all within `packages/client`.

**What changed** (`packages/client/src/api.ts`, `pageVisibility.ts`):

- **Escalating reconnect backoff.** `reconnectAfterMs` now escalates toward a **10-minute** cap (was a fixed 30s cap that Phoenix reset on every successful open). A new `trackConnectionStability()` watches `socket.onOpen`/`onClose` and keeps its own `unstableConnectionCount` that — unlike Phoenix's counter — survives brief "connect then immediately drop" cycles, and resets to 0 only once a connection stays up ≥30s. Identical to today's curve for the first ~6 attempts (1s → 30s), so genuine blips are unaffected.
- **`teardown()` always disconnects the socket**, even mid-reconnect (dropped the `isConnected()` guard). `disconnect()` is the only thing that cancels Phoenix's pending reconnect timer, so this stops a reauth (e.g. automatic token refresh) from leaking a socket that retries forever with soon-to-be-stale credentials.
- **Hidden tabs stop retrying.** `PageVisibilityManager` now parks a socket that is *mid-reconnect* when the page is hidden, not just a connected one, and resumes it when visible. Background zombie tabs drop to **zero** requests.
- **`online` resume trigger.** Reconnects promptly on the browser `online` event so recovery after a real network drop doesn't wait out the (now longer) backoff.

**Why "decay to slow" instead of "hard stop after N":** a hard cap silently kills realtime forever for users on flaky networks; a slow cadence self-heals after genuine outages while still cutting a dead-credential fleet's request volume by ~40–60× (and to zero for background tabs).

**Tunables reviewers may want to weigh in on:** the 10-minute terminal cadence (`SOCKET_RECONNECT_MAX_DELAY_MS`) and the 30s stability threshold (`STABLE_CONNECTION_THRESHOLD_MS`).

**Out of scope (follow-ups, noted on KNO-13857):** server-side accept-then-close-1000 on bad credentials (the only thing that helps *already-deployed* fleets, since client fixes require customers to upgrade + redeploy) and per-API-key connect rate limiting at the ALB.

### Checklist

Added coverage in `test/api.test.ts` (backoff escalation, ≥30s reset, clean-close no-op, teardown-while-disconnected, `online` single reconnect + listener cleanup) and `test/pageVisibility.test.ts` (mid-reconnect park/resume, never-activated no-op). Full client suite (582 tests), `type:check`, `lint`, and `format:check` all pass.

### DIFF
--- a/.changeset/reduce-ws-retry-storm.md
+++ b/.changeset/reduce-ws-retry-storm.md
@@ -1,0 +1,13 @@
+---
+"@knocklabs/client": patch
+---
+
+fix(KNO-13857): reduce websocket reconnect load when connections can't recover
+
+The Phoenix socket now escalates its reconnect backoff toward a 10-minute cap (previously a fixed 30s cap that Phoenix reset on every successful open), so a client that can never reconnect — for example after an API key rotation leaves stale credentials that get rejected on every upgrade — settles into a slow retry cadence instead of a tight loop. Backoff escalation persists through brief "connect then immediately drop" cycles and resets once a connection stays up for 30s.
+
+Additionally:
+
+- `teardown()` now always disconnects the socket (even mid-reconnect), so a reauth that replaces the client can't leak a socket that keeps retrying with stale credentials.
+- Hidden background tabs now stop retrying: a socket that is mid-reconnect when the page is hidden is disconnected, not just connected ones, and resumes when the page becomes visible again.
+- Reconnects promptly on the browser `online` event so recovery after a real network drop doesn't wait out the (now longer) backoff.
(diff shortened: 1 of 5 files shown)


kylemcd:
Superseded by #1031 (same changes, correct branch name `kyle-kno-13857-javascript-reduce-ws-retry-mechanism`).

